PR #205, opened 2020-10-28
ci: Install RACK on the Github Actions runner, test with Pytest

### LINKED ISSUE #204: ci: Run tests on the Github runner
John and I discussed a few options for how to run tests in CI via email. We settled on trying to install RACK and run tests natively on the Github runner, rather than building the RACK-box Docker container and running the tests against that, as I tried in #201.

I've started work on this branch: https://github.com/ge-high-assurance/RACK/compare/lb/ci-test-native.

langston-barrett:
Based on #201, but with tests run natively on the Actions runner. Fixes #204.

There is a speed-up over the strategy in #201 (first build the RACK-box image, then run it, and run tests against it).

- The installation step takes ~6:30 in Packer/Docker, and only ~3:40 on the Github Actions runner.
- We also don't have to cache, download, or build documentation files, saving a few seconds
- This method only waits for the SemTK services to start up once, not twice (saving ~1min or so)

It's possible that we would see more of a speed up if we used Docker's layering system more agressively, but this would come with an additional maintenance burden.

One remaining concern is whether the tests would always report the same results when run against the Github runner as they would when run on the official RACK-Box image. I plan to address this by running scheduled CI jobs that run the same test suite against the RACK-Box container (you can see the seeds of that in the `conftest.py` file in this PR).

### DIFF
--- a/.github/workflows/rack-box-continuous.yml
+++ b/.github/workflows/rack-box-continuous.yml
@@ -70,7 +70,7 @@ jobs:
         tar cfz RACK/rack-box/files/rack-cli.tar.gz RACK/cli/{setup-rack.sh,wheels}
 
     - name: Package RACK ontology and data
-      run: tar cfz RACK/rack-box/files/rack.tar.gz --exclude=.git --exclude=.github --exclude=assist --exclude=cli --exclude=rack-box --exclude=tools RACK
+      run: tar cfz RACK/rack-box/files/rack.tar.gz --exclude=.git --exclude=.github --exclude=assist --exclude=cli --exclude=rack-box --exclude=tools --exclude=tests RACK
 
     - name: Check out RACK wiki
       uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,76 @@
+name: Test RACK
+
+on:
+  pull_request:
+    branches: [ '*' ]
+  push:
+    branches: [ '*' ]
+    tags-ignore: [ '*' ]
+
+jobs:
+  test:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Check out RACK source
+      uses: actions/checkout@v2
+      with:
+        repository: ge-high-assurance/RACK
+        token: ${{ secrets.RACK_CI_PAT }}
+        path: RACK
+
+    - name: Download Fuseki distribution
+      if: steps.cache-fuseki.outputs.cache-hit != 'true'
+      run: curl -LSs https://apache.claz.org/jena/binaries/apache-jena-fuseki-3.16.0.tar.gz -o RACK/rack-box/files/fuseki.tar.gz
+
+    - name: Download SemTK distribution
+      run: curl -LSs https://github.com/ge-semtk/semtk/releases/download/v2.3.0-20201103/semtk-opensource-v2.3.0-20201103-dist.tar.gz -o RACK/rack-box/files/semtk.tar.gz
+
+    - name: Cache RACK CLI dependencies
+      uses: actions/cache@v2
+      with:
+        # This path is specific to Ubuntu
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('RACK/cli/requirements.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+          ${{ runner.os }}-
+
+    - name: Package RACK CLI
+      run: |
+        cd RACK/cli
+        python3 -m pip install --upgrade pip setuptools wheel
+        python3 -m pip wheel --wheel-dir=wheels -r requirements.txt
+        python3 -m pip wheel --wheel-dir=wheels .
+        cd ${{ github.workspace }}
+        tar cfz RACK/rack-box/files/rack-cli.tar.gz RACK/cli/{setup-rack.sh,wheels}
+
+    - name: Package RACK ontology and data
+      run: |
+        tar cfz RACK/rack-box/files/rack.tar.gz --exclude=.git --exclude=.github --exclude=assist --exclude=cli --exclude=rack-box --exclude=tests --exclude=tools RACK
+
+    - name: Download SemTK distribution
+      if: steps.cache-semtk.outputs.cache-hit != 'true'
+      run: curl -LSs 'https://oss.sonatype.org/service/local/artifact/maven/content?r=snapshots&g=com.ge.research.semtk&a=distribution&v=2.2.2-SNAPSHOT&c=bin&e=tar.gz' -o RACK/rack-box/files/semtk.tar.gz
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+
+    - name: Install RACK
+      run: |
+        cd RACK/rack-box
+        touch ./files/{documentation.html,index.html,style.css,systemctl3.py}
+        mkdir -p /tmp/files
+        for f in ./files/*; do
+          ln -s $(realpath ${f}) /tmp/files/$(basename ${f})
+        done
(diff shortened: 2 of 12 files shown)


rack-box/scripts/install.sh
```diff
@@ -128,7 +140,12 @@ done
 
 # Setup the RACK dataset using the RACK CLI
 
-cd /home/${USER}/RACK/RACK-Ontology/cli
+cd "/home/${USER}/RACK/RACK-Ontology/cli"
 # shellcheck disable=SC1091
 source venv/bin/activate
+# TODO(lb): Why is this necessary in Github Actions?
```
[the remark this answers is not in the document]
langston-barrett: Rebased onto that, testing now...

rack-box/scripts/install.sh
```diff
@@ -3,9 +3,18 @@
...
 cd /tmp/files
 
+# Helpers to bridge the Github Actions/Packer environments
+
+is_bin_on_path() { builtin type -P "${1}" &> /dev/null; }
+function do_sudo() {
+    if is_bin_on_path sudo; then
+      echo sudo
+    fi
+}
```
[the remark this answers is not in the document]
langston-barrett: Fixed.

.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
...
+        tar cfz RACK/rack-box/files/rack.tar.gz --exclude=.git --exclude=.github --exclude=rack-box --exclude=tests --exclude=tools RACK
+
+    - name: Download SemTK distribution
+      if: steps.cache-semtk.outputs.cache-hit != 'true'
+      run: curl -LSs 'https://oss.sonatype.org/service/local/artifact/maven/content?r=snapshots&g=com.ge.research.semtk&a=distribution&v=2.2.2-SNAPSHOT&c=bin&e=tar.gz' -o RACK/rack-box/files/semtk.tar.gz
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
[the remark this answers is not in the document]
langston-barrett: I'll try a branch without it and see what the effect is.

langston-barrett:
This needs a serious rebase.

.github/workflows/test.yml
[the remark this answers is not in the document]
langston-barrett: We resolved this elsewhere; it is indeed necessary.

langston-barrett:
Okay, I've rebased it, let's see what shakes loose...

langston-barrett:
Woah, this is an interesting error:
```
ERROR: Exception:
Traceback (most recent call last):
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/cli/base_command.py", line 228, in _main
    status = self.run(options, args)
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/cli/req_command.py", line 182, in wrapper
    return func(self, options, args)
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/commands/install.py", line 294, in run
    reqs = self.get_requirements(args, options, finder, session)
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/cli/req_command.py", line 334, in get_requirements
    for parsed_req in parse_requirements(
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/req/req_file.py", line 150, in parse_requirements
    for parsed_line in parser.parse(filename, constraint):
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/req/req_file.py", line 338, in parse
    for line in self._parse_and_recurse(filename, constraint):
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/req/req_file.py", line 343, in _parse_and_recurse
    for line in self._parse_file(filename, constraint):
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/req/req_file.py", line 376, in _parse_file
    _, content = get_file_content(
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/req/req_file.py", line 584, in get_file_content
    content = auto_decode(f.read())
  File "/home/runner/.local/lib/python3.8/site-packages/pip/_internal/utils/encoding.py", line 39, in auto_decode
    return data.decode(
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xac in position 11: invalid start byte
Error: Process completed with exit code 2.
```
From some online searching, this appears to be a bug in `pip`. However, we're already using the latest version of pip in CI:
```
Collecting pip
  Downloading pip-20.2.4-py2.py3-none-any.whl (1.5 MB)
```
https://pypi.org/project/pip/

Still looking for a workaroud...

.github/workflows/rack-box-continuous.yml
```diff
@@ -5,12 +5,7 @@
 
 name: RACK-in-a-Box continous build
 
-on:
-  pull_request:
-    branches: [ '*' ]
-  push:
-    branches: [ '*' ]
-    tags-ignore: [ '*' ]
+on: [push]
```
tuxji, line 8: I assume this change helps with not running multiple / redundant workflows, but how does it help?  Also, don't we still need tags-ignore?  My concern is that if we don't ignore tags when we publish a new RACK release, the new tag pushed by publishing the release will kick off all three of rack-box-continuous.yml, rack-box-release.yml, and rack-continuous.yml which still would be excessive.

.github/workflows/test.yml
```diff
@@ -0,0 +1,72 @@
...
+    - name: Install RACK
+      run: |
+        cd RACK/rack-box
+        touch ./files/{documentation.html,index.html,style.css,systemctl3.py}
+        mkdir -p /tmp/files
+        for f in ./files/*; do
+          ln -s $(realpath ${f}) /tmp/files/$(basename ${f})
+        done
+        export USER=$(whoami)
+        sudo --preserve-env=USER bash ./scripts/install.sh
```
tuxji, line 65: I'm 70% sure your `export USER=$(whoami)` is unnecessary.  If you open a new terminal and type `printenv USER` and `whoami`, both commands usually print the same string anyway.  Another, probably clearer, way to pass USER is to make scripts/install.sh look for an optional argument on the command line, e.g., `sudo bash scripts/install.sh $USER`.

rack-box/scripts/install.sh
```diff
@@ -3,7 +3,7 @@
 # Exit if anything goes wrong
 
 set -eo pipefail
-export USER=ubuntu
+export USER=${USER:-ubuntu}
```
tuxji, line 6: I think it would be clearer (and safer) to pass USER on the script's command line as an optional argument.  Please replace this line with `export USER=${1:-ubuntu}` which means if an argument is passed, use that argument instead of `ubuntu` as USER's value.

tuxji (changes requested):
+1 after you pass USER a better way.  I still don't like the extra quotes, but that's okay.

.github/workflows/rack-box-continuous.yml
langston-barrett: I made this change after my first push to this PR, which didn't trigger an associated CI build for some reason. Let me revert and see if that still happens...

langston-barrett: For context, see https://github.com/ge-high-assurance/RACK/issues/231

.github/workflows/test.yml
langston-barrett: Agreed, no idea why that didn't occur to me :smile: Thanks for the suggestion!

rack-box/scripts/install.sh
langston-barrett: Agreed, will do.

.github/workflows/rack-box-continuous.yml
langston-barrett: Yep, looks like it's working now. Reverted.

PR #216, opened 2020-11-02
Remove hasAttribute

glguy:


glguy:
@kquick or @Ptival could you check the change to assist?

Ptival:
@glguy 

The Prolog files in the ontology directory are actually generated via:
https://github.com/ge-high-assurance/RACK/blob/9b5b027072cc0df0b947b4ec403e6b0cf965a9c7/assist/bin/rack/write_ontology.pl#L179-L186

However, the files are otherwise out of date, as well as the instructions (since we have moved folder structures a bit).

Note that the OWL files in `OwlModel` still contain mentions of `hasAttribute`, so currently, regenerating prov-s.pl adds back `hasAttribute`.
Are we currently updating the OWL files alongside the SADL files? Or is this part of the current trouble with having OWL files checked in?

Anyway, I'm happy with you committing those changes to the Prolog file if you want, or to not update it, after which I can make regenerate it (as well as the other outdated files) and update the documentation where needed.

Ptival:
With the new changes, this seems good to me!

kquick:
@Ptival , should  the `write_ontology` process be done as part of CI?

Ptival:
@kquick oh that's a good idea!

langston-barrett:
> @Ptival , should the write_ontology process be done as part of CI?

> @kquick oh that's a good idea!

Let's make an issue!

glguy:
Covered by #217

PR #218, opened 2020-11-03
Tweak RACK CI workflows

tuxji:
Make it so that tag-only pushes will not trigger either rack-box-continuous.yml or rack-continuous.yaml workflows.  Now creating a new release will trigger only rack-box-release.yml, not all threee workflows.

Make rack-continuous.yaml a little more concise and remove superfluous shell: bash lines.

Streamline CI workflows and reduce duplication of lines by eliminating some caches that probably weren't saving us much time anyway.  The workflows were getting a little too long.  Also change our Docker Hub repository from interran/rack-box to gehighassurance/rack-box now that RACK is public.

In rack-box-continuous.yml and rack-box-release.yml, stop caching rack-cli.tar.gz, rack.tar.gz, documentation.html, index.html, style.css, and systemctl3.py.  Eliminate rack-box-release.yml's setup job (which was used only to populate some caches) and do all the work in its build job which will make it easier to compare rack-box-continuous.yml and rack-box-release.yml and keep both synchronized with each other.

In rack-continuous.yaml, tweak steps and comments slightly for better readability.

### DIFF
--- a/.github/workflows/rack-box-continuous.yml
+++ b/.github/workflows/rack-box-continuous.yml
@@ -7,7 +7,7 @@ name: RACK-in-a-Box continous build
 
 on:
   pull_request:
-    branches: [ master ]
+    branches: [ '*' ]
   push:
     branches: [ '*' ]
     tags-ignore: [ '*' ]
@@ -85,12 +85,12 @@ jobs:
         packer build -var version=dev rack-box-docker.json
 
     - name: Login to Docker Hub
-      if: github.base_ref == 'master'
+      if: github.base_ref == 'master' && github.event_name == 'push'
       uses: docker/login-action@v1
       with:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
 
     - name: Push rack-box image to Docker Hub
-      if: github.base_ref == 'master'
+      if: github.base_ref == 'master' && github.event_name == 'push'
       run: docker push gehighassurance/rack-box:dev
--- a/.github/workflows/rack-box-continuous.yml
+++ b/.github/workflows/rack-box-continuous.yml
@@ -1,11 +1,14 @@
-# Builds a Docker image upon each push to master:
+# Builds a Docker image upon every push (while ignoring tag-only pushes):
 #  - Caches files needed to build rack-box images
 #  - Builds rack-box:dev Docker image
 #  - Pushes rack-box:dev image to Docker Hub
 
 name: RACK-in-a-Box continous build
 
-on: [push]
+on:
+  push:
+    branches: [ '*' ]
+    tags-ignore: [ '*' ]
 
 jobs:
   build:
--- a/.github/workflows/rack-box-continuous.yml
+++ b/.github/workflows/rack-box-continuous.yml
@@ -6,6 +6,8 @@
 name: RACK-in-a-Box continous build
 
 on:
+  pull_request:
+    branches: [ master ]
   push:
     branches: [ '*' ]
     tags-ignore: [ '*' ]
--- a/.github/workflows/rack-box-continuous.yml
+++ b/.github/workflows/rack-box-continuous.yml
@@ -49,10 +49,16 @@ jobs:
       with:
         # This path is specific to Ubuntu
         path: ~/.cache/pip
+        # Look to see if there is a cache hit for the corresponding requirements file
         key: ${{ runner.os }}-pip-${{ hashFiles('RACK/cli/requirements.txt') }}
         restore-keys: |
           ${{ runner.os }}-pip-
           ${{ runner.os }}-
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
 
     - name: Package RACK CLI
       run: |
--- a/.github/workflows/rack-box-continuous.yml
+++ b/.github/workflows/rack-box-continuous.yml
@@ -63,6 +63,7 @@ jobs:
     - name: Package RACK CLI
       run: |
         cd RACK/cli
+        python3 -m pip install --upgrade pip setuptools wheel
         pip3 wheel --wheel-dir=wheels -r requirements.txt
         pip3 wheel --wheel-dir=wheels .
         cd ${{ github.workspace }}
--- a/.github/workflows/rack-box-continuous.yml
+++ b/.github/workflows/rack-box-continuous.yml
@@ -58,7 +58,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.x'
+        python-version: 3.8
 
     - name: Package RACK CLI
       run: |
--- a/.github/workflows/rack-box-continuous.yml
+++ b/.github/workflows/rack-box-continuous.yml
@@ -22,13 +22,6 @@ jobs:
         token: ${{ secrets.RACK_CI_PAT }}
         path: RACK
 
-    - name: Check out RACK wiki
-      uses: actions/checkout@v2
-      with:
-        repository: ge-high-assurance/RACK.wiki
-        token: ${{ secrets.RACK_CI_PAT }}
-        path: RACK.wiki
-
     - name: Cache Fuseki distribution
       uses: actions/cache@v2
       id: cache-fuseki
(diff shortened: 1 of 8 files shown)


langston-barrett:
I'm in meetings all day, but I can review by 5pm PST.

.github/workflows/rack-box-continuous.yml
```diff
@@ -1,11 +1,16 @@
...
 
 name: RACK-in-a-Box continous build
 
-on: [push]
+on:
+  pull_request:
+    branches: [ '*' ]
+  push:
+    branches: [ '*' ]
+    tags-ignore: [ '*' ]
```
tuxji, line 13: Now we will run our rack-box checks on both pull requests and pushes while ignoring tag-only pushes (hence avoiding starting redundant workflows when publishing a release).

.github/workflows/rack-box-continuous.yml
```diff
@@ -19,125 +24,80 @@ jobs:
...
-      uses: actions/cache@v2
-      id: cache-systemctl
-      with:
-        path: RACK/rack-box/files/systemctl3.py
-        key: systemctl-1.5.4260
-
-    - name: Download Fuseki distribution
-      if: steps.cache-fuseki.outputs.cache-hit != 'true'
-      run: curl -LSs https://apache.claz.org/jena/binaries/apache-jena-fuseki-3.16.0.tar.gz -o RACK/rack-box/files/fuseki.tar.gz
+        python-version: 3.8
```
tuxji, line 61: We have to specify python-version 3.8; we can't install a '3.x' (actually 3.9.0) wheel in our rack-box image because Ubuntu 20.04's native python is 3.8, not 3.9.

.github/workflows/rack-box-continuous.yml
```diff
@@ -19,125 +24,80 @@ jobs:
...
 
     - name: Package RACK CLI
-      if: steps.cache-rack-cli.outputs.cache-hit != 'true'
       run: |
         cd RACK/cli
+        python3 -m pip install --upgrade pip setuptools wheel
```
tuxji, line 66: I found out the cached python installation used by setup-python doesn't have wheel installed for some reason, so we need to install wheel explicitly.

.github/workflows/rack-box-continuous.yml
```diff
@@ -19,125 +24,80 @@ jobs:
...
       uses: docker/login-action@v1
       with:
         username: ${{ secrets.DOCKERHUB_USERNAME }}
         password: ${{ secrets.DOCKERHUB_TOKEN }}
 
     - name: Push rack-box image to Docker Hub
-      if: github.base_ref == 'master'
-      run: docker push interran/rack-box:dev
+      if: github.base_ref == 'master' && github.event_name == 'push'
+      run: docker push gehighassurance/rack-box:dev
```
tuxji, line 103: This allows us to check that pull requests can build rack-box images, but push rack-box images to Docker Hub only after pull requests are merged into the master branch.

.github/workflows/rack-box-release.yml
```diff
@@ -11,8 +11,29 @@ on:
...
+# virtualbox works only on macos (VT-x is not available: VERR_VMK_NO_VMX)
+        - builder: virtualbox
+          os: macos-latest
+          shell: bash
+
+    name: build ${{ matrix.builder }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: ${{ matrix.shell }}
```
tuxji, line 36: I've simplified the release workflow so that after you get past this part, it is almost identical to the continuous workflow thanks to eliminating the setup job, eliminating some of the caches, and running the same steps in both workflows up to the packer build.

.github/workflows/rack-box-release.yml
```diff
@@ -247,7 +144,7 @@ jobs:
 
     - name: Push rack-box image to Docker Hub
       if: matrix.builder == 'docker'
-      run: docker push interran/rack-box:${{ github.event.release.tag_name }}
+      run: docker push gehighassurance/rack-box:${{ github.event.release.tag_name }}
```
tuxji, line 147: Switching to gehighassurance/rack-box since RACK is now public.

.github/workflows/rack-continuous.yaml
```diff
@@ -1,58 +1,58 @@
...
 
 name: RACK continous lint
-on: [push]
+
+on:
+  pull_request:
+    branches: [ '*' ]
+  push:
+    branches: [ '*' ]
+    tags-ignore: [ '*' ]
```
tuxji, line 10: Likewise, now we will lint files on both pull requests and pushes while ignoring tag-only pushes.

.github/workflows/rack-continuous.yaml
```diff
@@ -1,58 +1,58 @@
...
 
 jobs:
   lint:
-    name: Lint shell scripts and the RACK CLI
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
tuxji, line 14: We're using Ubuntu 20.04 everywhere.

tuxji (commented):
Just explaining some of the changes I made.

.github/workflows/rack-box-continuous.yml
langston-barrett: Nice, avoiding redundant workflows sounds great. Do you know what the difference would be between the following two configurations?
```yaml
on:
  push:
    branches: [ '*' ]
```
vs
```yaml
on:
  pull_request:
    branches: [ '*' ]
  push:
    branches: [ '*' ]
```
Specifically, I'm curious if the second causes two builds for every push to a branch that also has an open PR.

.github/workflows/rack-box-continuous.yml
langston-barrett: Huh, that's good to know.

.github/workflows/rack-box-release.yml
```diff
@@ -23,214 +44,90 @@ jobs:
...
     - name: Install VirtualBox
       if: matrix.builder == 'virtualbox' && matrix.os == 'ubuntu-20.04'
       run: sudo apt-get install -y virtualbox
 
+    - name: Cache Ubuntu ISO
+      if: matrix.builder != 'docker'
+      uses: actions/cache@v2
+      id: cache-ubuntu
+      with:
+        path: RACK/rack-box/packer_cache/e30b34e4d3c536763868a42bb9d5b8f14af297b8.iso
```
langston-barrett, line 128: How big is this ISO? We discovered previously (#200) that the 5GB cache will start evicting the least recently used artifacts as soon as its full, so we should be sure that all of our rarely changing artifacts fit in that 5GB.

langston-barrett (approved):
This all looks great! Thanks for providing those comments on why different changes were made, it was all quite easy to look over and understand.

.github/workflows/rack-box-release.yml
tuxji: The ISO takes 914M bytes (in comparison, the semtk tarball, which we're not caching anymore since we're downloading it directly from a GitHub Release page now, takes a whopping 1.8G bytes).  However, I wouldn't be surprised if the ISO is no longer cached by the time we make another VirtualBox image a month from now.  The cache would save some time between a v3.9 release and a v4.0 release a few days later, though.  It's still a good question whether this cache is worth it (all I can say is that once I've run a packer build on my own computer and waited for packer to download the ISO, I'm happy packer caches the ISO in packer_cache to make the next packer build faster).

Regarding whether combining on: push: branches: with on: pull_request: branches: would cause two builds for every push to a branch that has an open PR, I made several pushes to this PR yesterday and I saw only one lint and one rack-box build started for these pushes.  There were two builds, but they weren't redundant builds (we wanted both the lint and the rack-box builds to run).

.github/workflows/rack-box-continuous.yml
langston-barrett: Following up on this: It looks like the semantics of this are that we only build on _pushes to branches that have an associated pull request_ which is a bit more restrictive than we had previously. In other words, the conditions in the two stanzas are ANDed together. 

@tuxji What are your thoughts on this configuration? I personally prefer running builds on pushes to _any_ branch, so that I can be sure my changes are passing CI before opening a pull request and notifying everyone who is watching the repo.

tuxji: I didn't know that the conditions in the two stanzas are AND'ed together.  I've seen other workflows that say:

`on: [pull_request, push]`

Can you find some GitHub documentation which states that `on: [pull_request, push]` is more restrictive than `on: pull_request` and `on: push` evaluated separately?  For that matter, I've also seen workflows which add `workflow_dispatch:` to these conditions and these workflows can be started from the GitHub GUI without making a git push or pull request as far as I understand.

langston-barrett: Huh, the documentation says

> ```yaml
> # Triggers the workflow on push or pull request events
> on: [push, pull_request]
> ```

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-using-a-list-of-events

I thought that I saw the AND behavior, but perhaps I was mistaken. Let's leave this for now and I'll gather more examples next time I see something that doesn't match my expectations.

PR #223, opened 2020-11-05
ci: Don't allow failure in ontology linting step

langston-barrett:
This was previously an allowed failure because it was failing on what we had checked in, now that it's no longer failing we can enforce it's checks going forward.

### DIFF
--- a/.github/workflows/rack-lint-continuous.yml
+++ b/.github/workflows/rack-lint-continuous.yml
@@ -43,13 +43,11 @@ jobs:
         mypy .
 
     - name: Lint RACK Ontology
-      continue-on-error: true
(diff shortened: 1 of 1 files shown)


langston-barrett:
Ah, looks like this step became out-of-date while it was in this "allowed failure" state:
```
ERROR: /home/runner/work/RACK/RACK/assist/bin/check:13: user:main directory_files/2: file `'models/TurnstileSystem/src'' does not exist
Error: Process completed with exit code 2.
```
Will fix.

langston-barrett:
```
.................................. Summary ...................................
No issues found
```
:tada:

PR #236, opened 2020-11-25
Fix typo in GrammaTech overlay: when->then

glguy:
@lucja-kot-grammatech Does this work for you?

Now the confidence values will correctly match the text fields.

```
	givenText (note "optional part that can be used to set certain conditions apart from other conditions in the ifText part") describes REQUIREMENT with values of type string.
	ifText (note "conditions under which this requirement holds") describes REQUIREMENT with values of type string.
	thenText (note "specifies what should hold by this requirement if the requirement conditions are met") describes REQUIREMENT with values of type string.
```


lucja-kot-grammatech:
Looks good - thanks!

PR #238, opened 2020-11-30
Run query nodegroups in testsuite

### LINKED ISSUE #195: Run sample query nodegroups in CI
We provide some sample nodegroups in the RACK distribution, we should run these in CI to ensure they stay up-to-date.

This could either mean:
 - Just run them and make sure there are no errors/exceptions
 - Compare their output to a file with the expected output

langston-barrett:
Fixes #195, though I plan to add a follow-up that compares the output against a file.

### DIFF
--- a/tests/test_nodegroups.py
+++ b/tests/test_nodegroups.py
@@ -1,5 +1,8 @@
 import pytest
 
+from glob import glob
+from pathlib import Path
+
 import semtk3
 import rack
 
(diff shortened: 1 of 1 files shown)


langston-barrett:
You can see that the tests are indeed running because there is a dot on this line JFYI: https://github.com/ge-high-assurance/RACK/pull/238/checks?check_run_id=1476798506#step:11:119

glguy (approved):
Running all the queries to check that they still work makes sense to me. This will catch nodegroups diverging from the ontology.

I wondered how reliably the nodegroup ID matches its filename part, and the answer I found was that it always exactly matches. No escaping is done.

Issue #193, opened 2020-10-23
Query sample data against the "invariant" nodegroups


langston-barrett:
I believe @cuddihyge mentioned that he and @glguy have developed two sets of nodegroups:

- One that checks for patterns that should be present in any valid data
- One that checks for patterns that should never occur in valid data

We should run these against the Turnstile data in CI.

@glguy can you point me to those nodegroups so I can get this started?

AbhaMoitra:
User will select which MIL-STD-881 appendix to load - so there will no name clash.

PR #239, opened 2020-12-02
Combine RACK workflow files

tuxji:
Strictly speaking, we don't need separate workflow files for jobs that
share the same trigger conditions.  Such jobs can go in the same
workflow file, so let's reduce the number of separate workflow files
to only two (continuous.yml for continuous lint, test, and build jobs
and release.yml for release build jobs).


tuxji:
@langston-barrett Can you take a look at the test job's output?  Especially https://github.com/ge-high-assurance/RACK/pull/239/checks?check_run_id=1488050269#step:11:119 which seems to show that a test failed, yet the test job itself still succeeded.  Was the test supposed to fail but the test job still succeed?

langston-barrett:
@tuxji The "x" there denotes an expected failure, fixing the underlying problem is on my TODO list (see #193).

Here's where the test is marked as an expected failure: https://github.com/ge-high-assurance/RACK/blob/fee08f6cb2da43a093f33f10ee9d6f56f917e62f/tests/test_invariant_nodegroups.py#L7

langston-barrett (approved):
LGTM, assuming that none of the individual steps changed.

Two questions:

- Are these run serially or in parallel?
- Is the filesystem shared between them? Would it be possible to share any resources (e.g. `pip` cache) between steps?

tuxji:
The individual steps haven't changed.  The jobs themselves have been combined or moved from multiple files to one file, but the following invariants still are true of the jobs:

The jobs run in parallel (even if defined together in the same file).

The jobs don't share the same filesystem (each job runs in its own virtual machine).

I don't know if the lint job can share the same pip cache as the other jobs because its requirement.txt file is different than the other jobs' requirement.txt file.  What is the reason for having two different files (cli/requirement.txt and cli/dev/requirement.txt)?

langston-barrett:
Thanks, that all makess sense.

> What is the reason for having two different files (cli/requirement.txt and cli/dev/requirement.txt)?

`cli/requirements.txt` lists the runtime dependencies of the `rack` package, whereas `dev/requirements.txt` lists tools that are used by *developers* of the `rack` package (e.g. linting tools).

PR #241, opened 2020-12-03
Reintroduce a few components into Turnstile

### LINKED ISSUE #240: Turnstile is missing COMPONENTs
Abha identified this bug in release testing. #60 introduced `COMPONENT`s into the Turnstile example and #126 added CSV files with them, but they're missing [on master](https://github.com/ge-high-assurance/RACK/blob/5f18025ca8464f01994b9b08e9d41bc2a9efb3ba/Turnstile-Ontology/02-Software/03-Implementation/CounterApplication.sadl).

glguy:
Fixes #240 

@russell-d-e @AbhaMoitra This commit reintroduces a small number of components back into turnstile so that the *Control Flow From Function* query has something to do. I didn't know how to use Dan's python code to generate the CSV or nodegroups and so I've just copied what we used to have to get things working quickly for the imminent release.

### DIFF
--- a/Turnstile-Ontology/02-Software/03-Implementation/CounterApplication.sadl
+++ b/Turnstile-Ontology/02-Software/03-Implementation/CounterApplication.sadl
@@ -48,6 +48,14 @@ input_o is a ObjectFile
     has file:createBy input_o_Compiling
     has fileFormat ElfFile.
 
+input_function is a COMPONENT
+    has identifier "input_function"
+    has dataInsertedBy Mockup
+    has componentType SOURCE_FUNCTION
+    has sw:name "input"
+    has valueType "void()"
+    has definedIn input_c.
+
 // output.o
 
 output_c is a SourceCode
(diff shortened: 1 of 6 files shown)


AbhaMoitra:
@russell-d-e : I had sent an email to some of the people regrading testing that I was doing. One item was "For “query Control Flow From Function” I did not get any results (it has run time constraint, but there are no values available). Are we missing some data? Should we remove this query to avoid confusion?"

So the changes here were introduced for that reason. Can you take a look?

langston-barrett (approved):
Thanks for fixing this!

russell-d-e:
> @russell-d-e : I had sent an email to some of the people regrading testing that I was doing. One item was "For “query Control Flow From Function” I did not get any results (it has run time constraint, but there are no values available). Are we missing some data? Should we remove this query to avoid confusion?"
> 
> So the changes here were introduced for that reason. Can you take a look?

I looked at this however I believe this is using some parts of the ontology that are not yet used in the Turnstile example as far as I know. I see that Langston has approved this pull request so I will assume things got corrected, otherwise we can talk tomorrow.

PR #242, opened 2020-12-03
Try to fix "Run rack-box tests"

glguy:


glguy:
@tuxji @langston-barrett  This makes the build succeed again. I canceled what looked like a duplicate task to avoid burning our minutes, but one everything looks like green checkmarks.

langston-barrett (approved):

tuxji:
OK, can Langston and you make a second pass over the changes (I know they were already merged, but still)?  I'm thinking that if the test actually requires semtk, then the same semtk line should be copied from cli/requirements.txt to tests/requirements.txt instead of running pip3 install -r cli/requirements.txt.  And the other line might not be needed because rack-box/scripts/install.sh already calls python3 -m pip install ./wheels/*.whl (which might be the same thing more or less?).  I'd be fine with any changes you want to make to the various python3 -m vs. pip3 etc. calls so that they're more consistent/correct in your opinion.

glguy:
@tuxji The right action might actually be to install that other requirements.txt because `test_invariant_nodegroups.py` relies on the `rack` module which comes from `pip install cli/.` and that requirements file has the requirements for `cli`. If we copy all of those requirements into `tests/requirements.txt` we're going to have to duplicate a lot of stuff.

langston-barrett:
> OK, can Langston and you make a second pass over the changes (I know they were already merged, but still)?

I can do this tomorrow.

> I'm thinking that if the test actually requires semtk, then the same semtk line should be copied from cli/requirements.txt to tests/requirements.txt instead of running pip3 install -r cli/requirements.txt.

In the original version, the test suite installed the CLI and its dependencies from the wheels we package: 

https://github.com/ge-high-assurance/RACK/pull/205/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R74

This is nice, because we're testing exactly what we ship (the packaged versions). I think that installing from the `cli/requirements.txt` (as done here) is also fine, but as you say, it's not clear why any action was necessary here at all:

> And the other line might not be needed because rack-box/scripts/install.sh already calls python3 -m pip install ./wheels/*.whl (which might be the same thing more or less?).

I can do some digging tomorrow to try and understand why that line doesn't obviate the ones added in this PR.

PR #252, opened 2020-12-22
Incorporate ontology review changes to PROV-S

glguy:


RACK-Ontology/ontology/PROV-S.sadl
```diff
@@ -10,41 +10,57 @@
...
+	invalidatedAtTime describes ENTITY with at most 1 value.
 
 COLLECTION (note "A collection is an entity that provides a structure to some constituents, which are themselves entities. These constituents are said to be member of the collections.")
 	is a type of ENTITY.
 	content (note "A collection is an entity that provides a structure to some constituents, which are themselves entities. These constituents are said to be member of the collections.") describes COLLECTION with values of type ENTITY.
 
 AGENT (note "An agent is something that bears some form of responsibility for an activity taking place, for the existence of an entity, or for another agent's activity.")
 	is a type of THING.
+	agentName describes AGENT with values of type string.
+	agentName describes AGENT with at most 1 value.
```
russell-d-e, line 54: for agentName should we consider having multiple agentNames for situations like if they wanted to capture first/last name or aliases for things like the same person had a name change (i.e. maiden names)

russell-d-e (commented):
had comment on the cardinality for agentName

AbhaMoitra:
@davearcher - you are the reviewer and since we agreed on this, can you review / approve this request AFTER the conflict is resolved.

RACK-Ontology/ontology/PROV-S.sadl
davearcher: Proposing to drop agentName altogether in favor of using Title in the base class.

davearcher:
I think this one is good to go, after agentName is removed by @glguy

PR #249, opened 2020-12-18
Assist file updates

kquick:
This PR updates the ASSIST tooling:
* Accommodations for changes in RACK ontology (esp. new FILE module)
* Additional recognizers for various fields
* Some bug-fixes in the tooling itself
* Ability to generate instances based on other instances (e.g. FILE_HASH instances for a FILE instance)
* Support for user-provided recognizers

PR #253, opened 2020-12-22
Turnstile assist

kquick:
This updates the Turnstile example to use ASSIST for the software structure data generation and ingestion.

This can be tested using the same technique described in https://github.com/ge-high-assurance/RACK/pull/249#issuecomment-748243498, although the `ingest_data` command should be supplemented with `-r Turnstile-Ontology/02-Software/turnstile-ingest.rack` before the positional parameters.


glguy:
Just looking at the diff we seem to have a lot few triples than before. Is this before the triples are now all generated instead of committed into the repository?

kquick:
I assume you are referring to OWL files generated when you loaded Eclipse SADL (I did not include these changes in the PR, feel free to add them).  

I would expect fewer because the software structure instance data is generated and loaded directly by ASSIST.  You can specify a `-o foo.owl` to the `ingest_data` tool which will write a local OWL file, but it will be very difficult to compare to the SADL-generated OWL files because there is no normalization in OWL on-disk format.

AbhaMoitra:
@kquick - thanks

glguy (approved):

russell-d-e (approved):

PR #254, opened 2020-12-22
Incorporate ontology-review changes to ANALYSIS

glguy:


RACK-Ontology/ontology/ANALYSIS.sadl
```diff
@@ -10,32 +10,44 @@
...
+  
+uri "http://arcos.rack/ANALYSIS" alias ^An.
 import "http://arcos.rack/PROV-S".
+import "http://arcos.rack/DOCUMENT".
 
 ANALYSIS_REPORT
-	(note "Output of an ANALYSIS")
-	is a type of ENTITY.
+    (note "A document that is the output of an ANALYSIS")
+    is a type of REPORT.
```
russell-d-e, line 29: I don't think that the note should be changed and  this should be a subclass of the Document Report. I would actually expect a Document Report to contain a collection of ANALYSIS_REPORTs, which describe a single analysis, An example may be a code coverage report where each line of code is is tracked as hit or not. Then each of these individual reports would be aggregated into a full REPORT document.  The ANALYSIS_REPORT is intended what is produced by the ANALYSIS, similar to a REVIEW and a REVIEW_LOG.

Maybe we should change the ANALYSIS_REPORT to some else like a ANALYSIS_STATEMENT or something along those lines. This would also allow us to remove the document import.

russell-d-e (changes requested):

RACK-Ontology/ontology/ANALYSIS.sadl
glguy: OK, so I think the distinction is that we're thinking of the analysis output as simple computer-generated output and not a curated, versioned document kind of thing

russell-d-e: Yes.  Maybe ANALYSIS_OUTPUT would be the better term.

PR #257, opened 2020-12-23
Incorporate ontology review for FILE

glguy:


RACK-Ontology/ontology/FILE.sadl
```diff
@@ -10,29 +10,44 @@
...
     createBy (note "FILE_CREATION ACTIVITY that created this file") describes FILE with values of type FILE_CREATION.
+    createBy describes FILE with at most 1 value.
     createBy is a type of wasGeneratedBy.
     
     fileHash (note "Hash of the file contents") describes FILE with values of type FILE_HASH.
 
 FORMAT
     (note "The data format of a FILE")
     is a type of THING.
+ADA is a FORMAT with identifier "ADA".
```
russell-d-e, line 48: I don't think we should define instances in here seems that is something that should be loaded into the base ontology

russell-d-e (changes requested):

RACK-Ontology/ontology/FILE.sadl
glguy: Do we need a Boeing-Ontology for things common across all the various teams to have access to?

glguy:
I think this PR is a straight forward change at this point and is worth merging.

PR #259, opened 2020-12-23
Incorporate ontology review for REQUIREMENTS

glguy:


RACK-Ontology/ontology/REQUIREMENTS.sadl
```diff
@@ -10,62 +10,73 @@
...
-	text (note "Description of the REQUIREMENT") describes REQUIREMENT with values of type string.
+	text (note "Description of the REQUIREMENT") describes REQUIREMENT with a single value of type string.
 	// The "text" string can optionally be subdivided into sub-parts : givenText, ifText, thenText
 	// This corresponds to how rules are defined in SADL and note that in a SADL rule whatever is in
 	// givenText can simply be included in the ifText part.
 	// Note that modulo naming changes this structure also matches the structure
 	// that GrammaTech had which was : Given, When, Then parts.
 	// The structure of requirements in ASSERT subsume this structure of requirements. 
 	givenText (note "optional part that can be used to set certain conditions apart from other conditions in the ifText part") describes REQUIREMENT with values of type string.
+	givenText describes REQUIREMENT with at most 1 value.
```
russell-d-e, line 36: I don't think this should have a cardinality restriction. there is no reason that you could not have multiple. 

Also in a common should the there be a new ENTITY defined for GIVEN_TEXT, in general it seems like the ravage should not be a string for anything outside the THING class. since how would TA3 be able to use/query for that. Seems like we should use a specific class, or this should be moved to the supplemental ontology

RACK-Ontology/ontology/REQUIREMENTS.sadl
```diff
@@ -10,62 +10,73 @@
...
 	ifText (note "conditions under which this requirement holds") describes REQUIREMENT with values of type string.
+	ifText describes REQUIREMENT with at most 1 value.
```
russell-d-e, line 38: I don't think this should have a cardinality restriction. there is no reason that you could not have multiple. 

Also in a common should the there be a new ENTITY defined for GIVEN_TEXT, in general it seems like the ravage should not be a string for anything outside the THING class. since how would TA3 be able to use/query for that. Seems like we should use a specific class, or this should be moved to the supplemental ontology

RACK-Ontology/ontology/REQUIREMENTS.sadl
```diff
@@ -10,62 +10,73 @@
...
 	thenText (note "specifies what should hold by this requirement if the requirement conditions are met") describes REQUIREMENT with values of type string.
+	thenText describes REQUIREMENT with at least 1 value.
```
russell-d-e, line 40: I don't think this should have a cardinality restriction. there is no reason that you could not have multiple. 

Also in a common should the there be a new ENTITY defined for GIVEN_TEXT, in general it seems like the ravage should not be a string for anything outside the THING class. since how would TA3 be able to use/query for that. Seems like we should use a specific class, or this should be moved to the supplemental ontology

RACK-Ontology/ontology/REQUIREMENTS.sadl
```diff
@@ -10,62 +10,73 @@
...
 	satisfies (note "Parent ENTITY(s) (e.g. REQUIREMENT) that this REQUIREMENT is derived from") describes REQUIREMENT with values of type ENTITY.
-	satisfies is a type of wasDerivedFrom.
-
-	mitigates (note "ENTITY(s) (e.g. HAZARD) that is being mitigated by this REQUIREMENT") describes REQUIREMENT with values of type ENTITY.
-	mitigates is a type of wasDerivedFrom.
+	satisfies is a type of wasImpactedBy.
 
-	createdBy (note "ACTIVITY(s) (e.g. REQUIREMENT_DEVELOPMENT) that created this REQUIREMENT") describes REQUIREMENT with values of type ACTIVITY.
-	createdBy is a type of wasGeneratedBy.
+	Rq:mitigates (note "ENTITY(s) (e.g. HAZARD) that is being mitigated by this REQUIREMENT") describes REQUIREMENT with values of type HAZARD.
```
russell-d-e, line 48: Why are we limiting this to HAZARD?  What if we are wanting to mitigate something else like a failure condition or something

RACK-Ontology/ontology/REQUIREMENTS.sadl
```diff
@@ -10,62 +10,73 @@
...
 
+	Rq:createdBy (note "ACTIVITY(s) (e.g. REQUIREMENT_DEVELOPMENT) that created this REQUIREMENT") describes REQUIREMENT with values of type ACTIVITY.
+	Rq:createdBy is a type of wasGeneratedBy.
 
 DATA_DICTIONARY_TERM
 	(note "Defines some conceptual data that may be collected somewhere and used somewhere else")
 	is a type of ENTITY.
 
 	text (note "Description of the term") describes DATA_DICTIONARY_TERM with values of type string.
+	text describes DATA_DICTIONARY_TERM with at most 1 value.
```
russell-d-e, line 59: I think that this property can be removed as I think it can be covered by the proerties already defined for THING (title, description, identifier).  This may force some updateds to the turnstile example to replace text with description

russell-d-e (changes requested):

AbhaMoitra:
@davearcher - can you approve this as you were the selected reviewer. This PR is the one we decided to merge.

RACK-Ontology/ontology/REQUIREMENTS.sadl
davearcher: I'm fine with loosening the cardinality restriction. Not sure by we'd define an entity class for givenText, though - it seems to clearly be an attribute of the thing rather than another thing.

RACK-Ontology/ontology/REQUIREMENTS.sadl
davearcher: I'm fine with relaxing the cardinality restriction. Problem I see is that it's not clear whether the ifTexts are conjunctive or disjunctive if we do that. Is it for example that ALL of the ifs apply? or that ANY of the ifs apply? If Dan has a clean answer for that question, we can probably close this one.

PR #260, opened 2020-12-23
Incorporate ontology review for REVIEW

glguy:
Note that while Dave proposed removing `governedBy` because it was already covered by properties of REQUIREMENT that in Turnstile governedBy was used to point to a `SPECIFICATION`. We'll need to either leave it in or figure out some other way to express what Turnstile says.

### DIFF
--- a/RACK-Ontology/ontology/REVIEW.sadl
+++ b/RACK-Ontology/ontology/REVIEW.sadl
@@ -10,15 +10,24 @@
  * of the Defense Advanced Research Projects Agency (DARPA).
  */
 
+/************** edit history *****************
+  *
+  * dwa - 11 Nov 2020 - reduced cardinality of several
+  * dwa - 11 Nov 2020 - removed incorrect associations
+  * dwa - 11 Nov 2020 - removed unnecessary subclassing
+  *
+  *********************************************/
+
 uri "http://arcos.rack/REVIEW" alias Rv.
 import "http://arcos.rack/PROV-S".
+import "http://arcos.rack/DOCUMENT".
 
 REVIEW_LOG
 	(note "Outcome of a REVIEW ACTIVITY")
-	is a type of ENTITY.
+	is a type of REPORT.
 
-	reviews (note "ENTITY(s) that the review was performed on") describes REVIEW_LOG with values of type ENTITY.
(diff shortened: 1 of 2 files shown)


RACK-Ontology/ontology/REVIEW.sadl
```diff
@@ -10,15 +10,24 @@
...
+  *********************************************/
+
 uri "http://arcos.rack/REVIEW" alias Rv.
 import "http://arcos.rack/PROV-S".
+import "http://arcos.rack/DOCUMENT".
 
 REVIEW_LOG
 	(note "Outcome of a REVIEW ACTIVITY")
-	is a type of ENTITY.
+	is a type of REPORT.
```
russell-d-e, line 27: I don't think a  REVIEW_LOG should be a subclass of REPORT, a REPORT would be a collection of  REVIEW_LOGs. For example a REVIEW_LOG would be generated for each REQUIREMENT, then the REPORT would aggregate the REVIEW_LOG items for all the REQUIREMENTs associated with a particular SYSTEM.  This would also allow us to remove the DOCUMENT import

RACK-Ontology/ontology/REVIEW.sadl
```diff
@@ -29,14 +38,14 @@ REVIEW_STATE (note "Life-cycle status of a review log")
...
+ReviseWithoutReview is a REVIEW_STATE with identifier "Revise Without Review".
 ReviseWithReview    is a REVIEW_STATE with identifier "Revise With Review".
 
 REVIEW
-	(note "Code review ACTIVITY")
+	(note "Artifact review ACTIVITY")
 	is a type of ACTIVITY.
 
-	author (note "AGENT(s) putting items up for review") describes REVIEW with values of type AGENT.
+	author (note "AGENT(s) putting items up for review") describes REVIEW with a single value of type AGENT.
```
russell-d-e, line 48: while in the norm I would agree the cardinality would be 1, I am not sure I this would always be the case.  I would suggest to just leave this cardinality unrestricted.  If we do keep the restriction then the note should be updated to remove the "(s)" from AGENT

russell-d-e (changes requested):

PR #272, opened 2021-01-11
cli: emit a warning guessing command to correct mistake

Ptival:
When one passes the wrong file to the command-line argument, we get a pretty
elaborate error message.  This adds a one-line warning suggesting a command
that would likely work given the fields in the input file.

That is, if the fields in the input file correspond to fields we expect in some
known command, this suggests the user to try that command instead.

image

### DIFF
--- a/cli/bin/rack
+++ b/cli/bin/rack
@@ -24,6 +24,7 @@ import yaml
 from jsonschema import ValidationError
 
 from rack import CustomFormatter, get_argument_parser
+from rack import cliMethod, CLIMethod, INGEST_CSV_CONFIG_SCHEMA, INGEST_OWL_CONFIG_SCHEMA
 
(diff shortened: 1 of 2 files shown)


glguy (approved):
This is a great idea. Is it possible that the user would have been running the suggested command already but still generate this suggestion due to a different validation error?

Ptival:
Indeed I think it may be possible (though likely quite rare).

It's possible to check, though it is slightly awkward, since there is currently no "state" keeping track of what command was issued (the parser merely dispatches to the appropriate handler), so I think we'd end up wanting some sort of global state since the information needs to survive the exception?

What do you think?

Ptival:
Now recording which CLI method was called upon dispatch, and checking that the user did not use the recommended command before issuing the warning.

PR #278, opened 2021-01-16
Load all of our process documents into separate data graphs

glguy:
This pull request pull in all of our process documents but into separately named graphs. It them specifically pulls in the DO-178C graph when importing the Turnstile example.


Ptival (commented):
@glguy is it expected that the `SadlBaseModel.owl` and `SadlListModel.owl` files are included in this PR?

The rest looks good to me!

PR #279, opened 2021-01-21
updated names of all returned results so that they follow a convention

### LINKED ISSUE #270: Update nodegroup queries so that they match the ontology/CDR.
See for example the "query Hazard Structure" in the nodegroup.

kityansiu:
This resolves #270.


glguy:
5 of these nodegroups generated errors when I tried to query with them:

- query Interface structure
- query Hazard structure
- query System Structure
- query Requirements with failed test result
- query Requirements without passed test

I'm going to take a look next to see what went wrong.

glguy:
To get those 5 queries to work I had to open them in SemTK and re-save them (without any modification). I don't know why though.

kityansiu:
I did not test with cli, so it's a good thing you ran into this. When you re-saved those 5 nodegroups, were there any differences in the .json files?

glguy:
> I did not test with cli, so it's a good thing you ran into this. When you re-saved those 5 nodegroups, were there any differences in the .json files?

They were quite different, actually, as far as things coming out in much different orders. It makes me wonder if we're using different versions of SemTK. If you want I can resave those and push that change to this PR.

cuddihyge:
I would be happy to review the before/after and make sure nothing too fishy is going on.

There has been a new version of semtk.  And there will be one next week also.  The nodegroup files haven't changed though.   It is possible that a little bit of reconciliation against the model occurs when you load it, but it isn't entirely clear to me what happened.

kityansiu:
> 
> 
> > I did not test with cli, so it's a good thing you ran into this. When you re-saved those 5 nodegroups, were there any differences in the .json files?
> 
> They were quite different, actually, as far as things coming out in much different orders. It makes me wonder if we're using different versions of SemTK. If you want I can resave those and push that change to this PR.

Yes, you're right about the order being quite different. When I was updating all of the query nodegroups, I noticed the difference in order. So what I did was manually edit the last GitHub versions in the places I knew where the header names were defined. I tested all the updated .json files in SPARQLgraph. Something else must be different that I didn't manually edit.

This brings up a question. Should we overwrite all the query nodegroup .json files, and lose the ability see the diff between this version and last because the ordering will be all different?

glguy:
It looks like this is the relevant portion.

```diff
--- /dev/fd/11	2021-01-22 14:21:51.000000000 -0800
+++ /dev/fd/12	2021-01-22 14:21:51.000000000 -0800
@@ -42,11 +42,7 @@
   "sNodeGroup": {
     "limit": 0,
     "offset": 0,
-    "orderBy": [
-      {
-        "sparqlID": "?hazard"
-      }
-    ],
+    "orderBy": [],
```

glguy (approved):
I've removed the troublesome orderBy clause and now these queries work from CLI

AbhaMoitra:
I am a bit confused. Do we know why we had a "orderBy" that needed to be dropped. I thought @kityansiu had hand edited to just change the headers. 

Also, I tried running "query Requirements with failed test result.json" but when I tried loading it, I got the following error.

image

I am running RACK v4.1 without the updated SemTK.

kityansiu:
> 
> 
> I am a bit confused. Do we know why we had a "orderBy" that needed to be dropped. I thought @kityansiu had hand edited to just change the headers.
> 
> Also, I tried running "query Requirements with failed test result.json" but when I tried loading it, I got the following error.
> 
> image
> 
> I am running RACK v4.1 without the updated SemTK.

@AbhaMoitra, test with the dev version of RACK. The error you're seeing is because of range tightening from the ontology review. (Was also commented on in #280.) So when merged into the Master, these query nodegroups should work with the dev.

cuddihyge:
If you are troubleshooting, please note that there is an outside chance that a multiple failure occurred including two properties in the model with the same name on the same class, and a SemTK bug that didn't handle it.   This MIGHT or MIGHT NOT contribute.

In any event we are pushing a semtk release 20210125.  Please wait for that one and let's make sure we're on the same version before we dive into this.

glguy:
> I am a bit confused. Do we know why we had a "orderBy" that needed to be dropped. I thought @kityansiu had hand edited to just change the headers.

I don't know why this change was needed; it's just the change that happened when I saved the nodegroup using SemTK.

cuddihyge:
Note that semtk release v2.0.3-20210125 is now built, and it **should** be automatically downloaded into the latest RACK docker.

@kityansiu I would like to try to reproduce the disappearing "ORDER BY" phenomenon.

kityansiu:
> 
> @kityansiu I would like to try to reproduce the disappearing "ORDER BY" phenomenon.

@cuddihyge, here's an example of this phenomenon. In your last commit [8abdf1f](https://github.com/ge-high-assurance/RACK/pull/279/commits/8abdf1f95fedc7c8beed5198f26cd5c524a7fac6), see how the original file, line 57, had an empty clause in orderBy. The updated file, line 367, has orderBy filled in. Do you recognize this as a SemTK change? If it is, perhaps 
@glguy, the CLI needs to be updated?

AbhaMoitra:
Update on what I ran. I am using the dev version of RACK and 2 of the 3 queries run fine. The third one - "query Requirements without passes test.json" is to be run with the new semtk release which I do not have currently.

AbhaMoitra:
I ran all 3 queries on the updated RACK/SemTk of 1/25 or 1/26 and there were no errors but the union query did not return any results. I added this info to #280 as well.

AbhaMoitra (approved):

AbhaMoitra:
With the dev build of 1/28, all 3 queries ran fine.

PR #282, opened 2021-01-28
Added dateTime for a single SoftwareUnitTestResult

AbhaMoitra:
Added dateTime to just UTR-1-1-1 to illustrate syntax and test RACK workflow. This works with dev version of RACK dated 1/28/2021/


RACK-Ontology/OwlModels/SACM-S.owl
```diff
@@ -18,12 +18,12 @@
     <rdfs:comment xml:lang="en">This ontology was created from a SADL file 'SACM-S.sadl' and should not be directly edited.</rdfs:comment>
   </owl:Ontology>
   <owl:Class rdf:ID="ASSERTED_CONTEXT">
-    <rdfs:comment xml:lang="en">Asserted Context can be used to declare that the artifact cited by an ArtifactReference(s) provides the context for the
+    <rdfs:comment xml:lang="en">Asserted Context can be used to declare that the artifact cited by an ArtifactReference(s) provides the context for the&#xD;
 interpretation and scoping of a Claim or ArgumentReasoning element.</rdfs:comment>
```
kquick, line 22: What is this change?  It doesn't seem to correspond to a change in any .sadl file, and it doesn't seem like a change we want?

RACK-Ontology/OwlModels/SACM-S.owl
```diff
@@ -18,12 +18,12 @@
...
     <rdfs:subClassOf rdf:resource="PROV-S#ENTITY"/>
   </owl:Class>
   <owl:Class rdf:ID="ASSERTED_INFERENCE">
-    <rdfs:comment xml:lang="en">Asserted Inference association records the inference that a user declares to exist between one or more Assertion (premise) and
+    <rdfs:comment xml:lang="en">Asserted Inference association records the inference that a user declares to exist between one or more Assertion (premise) and&#xD;
```
kquick, line 26: Another odd character introduction.

RACK-Ontology/OwlModels/SACM-S.owl
```diff
@@ -32,7 +32,7 @@ another Assertion (conclusion).</rdfs:comment>
     <rdfs:subClassOf rdf:resource="PROV-S#ENTITY"/>
   </owl:Class>
   <owl:Class rdf:ID="ARGUMENT_PACKAGE">
-    <rdfs:comment xml:lang="en">Argument Package is the containing element for a structured argument represented using the SACM Argumentation
+    <rdfs:comment xml:lang="en">Argument Package is the containing element for a structured argument represented using the SACM Argumentation&#xD;
```
kquick, line 35: Another odd character introduction.

Turnstile-Ontology/99-Utils/Data/SoftwareUnitTestResult.csv
```diff
@@ -1,10 +1,10 @@
...
-"Passed","SoftwareUnitTestRun1","UTC-1-1","UTR-1-1-1"
+"identifier_TEST_STATUS","generatedAtTime_SoftwareUnitTestResult","identifier_SoftwareUnitTestExecution","identifier_SoftwareUnitTest","identifier_SoftwareUnitTestResult"
+"Failed",null,"SoftwareUnitTestRun2","UTC-1-4","UTR-1-4-2"
+"Passed",null,"SoftwareUnitTestRun2","UTC-1-3","UTR-1-3-2"
+"Passed",null,"SoftwareUnitTestRun2","UTC-1-2","UTR-1-2-2"
+"Passed",null,"SoftwareUnitTestRun2","UTC-1-1","UTR-1-1-2"
+"Failed",null,"SoftwareUnitTestRun1","UTC-1-4","UTR-1-4-1"
+"Passed",null,"SoftwareUnitTestRun1","UTC-1-3","UTR-1-3-1"
+"Passed",null,"SoftwareUnitTestRun1","UTC-1-2","UTR-1-2-1"
+"Passed","Wed Mar 22 20:00:00 EDT 2017","SoftwareUnitTestRun1","UTC-1-1","UTR-1-1-1"
```
kquick, line 9: This timestamp doesn't seem to be the same as the one in CounterApplicationUnitTesting.sadl, even allowing for timezone conversions?

AbhaMoitra:
@kquick : good catches! Not sure why the weird characters got introduced. Looking into the dateTime issue.

glguy:
I'm working on update this PR against ontology changes. I'll push the updates soon.

There's definitely something wrong with the way the date times are being rendered to CSV, however. We should fix that, but it probably doesn't need to block this from being merged.

glguy:
We'll want to re-run CSV generation with the ASSIST generated data in place. I'm not exactly sure how that works, yet.

glguy:
OK, I found the issue. Don't use `T` in your dateTime literals, SADL doesn't like that. Use a *space* instead.

AbhaMoitra:
@glguy : Thank you very much! Here is what I see - want to confirm it with you.
In SADL I have has generatedAtTime "2017-03-23 10:03:16"
In OWL I see (note -4:00)
<j.0:generatedAtTime rdf:datatype="http://www.w3.org/2001/XMLSchema#dateTime"
    >2017-03-23T10:03:16-04:00</j.0:generatedAtTime>
In csv I see Thu Mar 23 10:03:16 EDT 2017
In SemTK I see 2017-03-23T10:03:16-04:00

glguy:
@AbhaMoitra as you pointed out the example wasn't specifying a timezone. Unfortunately SADL is rendering the results out in the local timezone, but I've updated the input to use UTC explicitly. We'll want people to specify timezones in their inputs times so that the semantics of their files don't depend on the current location, so this will make for an even better example.

russell-d-e (approved):

PR #287, opened 2021-02-03
cli: minimal retry

Ptival:
This adds a retry behavior to wrapped SemTK commands that may fail due to
temporary network conditions.

Questions:
- [ ] Is this enough?
- [ ] Should some commands **not** be retried?
- [ ] Should the delay be user-controlled?
- [ ] Should the decision of allowing retries be user-controlled?

### DIFF
--- a/cli/rack/__init__.py
+++ b/cli/rack/__init__.py
@@ -153,7 +153,6 @@ def retry_on_exception(func: Decoratee) -> Decoratee:
             print(f"This command raised an exception, will attempt retrying in {RETRY_DELAY} seconds... Press ^C to abort.")
             time.sleep(RETRY_DELAY)
             return wrapper(*args, **kwargs)
-        return result
     return cast(Decoratee, wrapper)
 
 # Bug https://github.com/PyCQA/pylint/issues/1953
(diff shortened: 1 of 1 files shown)


cli/rack/__init__.py
```diff
@@ -142,6 +143,19 @@ def format(self, record: logging.LogRecord) -> str:
...
+    RETRY_DELAY = 3
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        try:
+            return func(*args, **kwargs)
+        except Exception as e:
+            print(e)
+            print(f"This command raised an exception, will attempt retrying in {RETRY_DELAY} seconds... Press ^C to abort.")
+            time.sleep(RETRY_DELAY)
+            return wrapper(*args, **kwargs)
+        return result
```
glguy, line 156:
```suggestion
```

glguy:
I think it might be good to not show the error message the first time that something else to retry. We're trying to give the impression that the service is more reliable.

It would be good to know (and I'm not sure if it's true or not) that it's safe to just repeat all the commands and that they won't leave behind an incomplete change in some cases. @cuddihyge is it always safe to run rerun imports of CSV, owl, and nodegroups on failure? Do these rollback all changes on error?

cuddihyge:
Rerunning an import is potentially unsafe.   Nodegroups and owl should be fine.

Imports tell you whether they were partially completed.  They do lots of internal retries so it is less likely to happen.  If we get a REST failure checking the status of an import we certainly don't want to start over.

I think we should try pushing the retries down to the lowest level of semtk-python3.  Not up in CLI.
I also think we need @tuxji to help us give folks a way to check whether their docker container is running ok.

PR #291, opened 2021-02-05
Vr/ada

Ptival:
I have updated the Ada analysis code to reflect recent changes to the ontology, though I don't have a decent tool to check whether the generated Turtle indeed follows the current state of the OWL files.

This could be a good point to merge into master.


glguy (approved):
This works when I run it against the Boeing source code

PR #295, opened 2021-02-10
Import SACM-LM overlay

glguy:
This is the almost unmodified SADL we got from LM


AbhaMoitra:
* When I pulled, there was no .project file
* In SACM-S we had defined ARGUMENT_PACKAGE and LM is defining scms:ARGUMENT_PACKAGE. There are other such instances. Should we remove something in our core ontology to avoid confusion?
* I do not like the style used for properties as it is not in synch with our style. For example they use HAS_CONTENT as a property; I would prefer it to be hasContent. There are other instance of this style in the file. Is this something we want to change?
* We had decided to use SADL setting "Translate multiple-class domain or range as union class". As a result of this the following 3 SADL lines
      HAS_ELEMENT describes TerminologyPackage with values of type TerminologyElement.
      HAS_ELEMENT describes TerminologyGroup with values of type TerminologyElement.
      HAS_ELEMENT describes sacms:ARGUMENT_PACKAGE with values of type ArgumentationElement.

gets translated to the following OWL snippet which is probably not what they would want. A solution is to have the first 2 lines in one SADL file and the 3rd in a different file - I have not tried it though.

image

glguy:
The .project file must have been ignored when I was making the commit. I actually have one based on the other projects locally; I’ll fix.

glguy:
I'm definitely on board with cleaning this up to match the style. I wanted to get the original request incorporated first and as early as possible.

PR #300, opened 2021-02-13
ada: use ScrapingToolKit as output

Ptival:
This large refactoring allows us to use the ScrapingToolKit to output
information from the analysis.

In order to not completely scrape away the Turtle output, and allow us
to later make different decisions, I tried to have a layer between the
analysis computing the output and the backend outputting it.

Because we were manipulating a lot of strings, including strings to be
used as keys in Python, strings to be used as keys in the ontology data,
strings to be used as names, ..., I also added 'NewType's that prevent
us from getting these mixed up, at the price of a bit of wrapping or
casting in some places.

Currently the output to Turtle is preserved behind a 'DEBUG_TURTLE'
flag, but it may make sense to either drop it entirely, or put it behind
a command-line argument later.

### DIFF
--- a/ada/ontology.py
+++ b/ada/ontology.py
@@ -4,18 +4,42 @@ from __future__ import annotations
 
 __copyright__ = "Copyright (c) 2020, Galois, Inc."
 
-from enum import Enum
-from typing import List, Optional
+from typing import NewType, Optional, Set
+# from typing_extensions import TypedDict
 from rdflib import Graph, Literal, Namespace, URIRef
 
-FILE = Namespace("http://arcos.rack/FILE#")
-PROV_S = Namespace("http://arcos.rack/PROV-S#")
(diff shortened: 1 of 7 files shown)


glguy:
I'm seeing

```
COMPONENT_TYPE1.csv  COMPONENT_TYPE2.csv  FILE1.csv  FILE2.csv  FORMAT1.csv  FORMAT2.csv  import.yaml
```

generated but I don't see any software components being generated.

Ptival:
There is a `RACK-DATA.xml` that contains `SWCOMPONENT`s, indeed they don't appear in any CSV file.

glguy (approved):
This code is working well for us. We can work on building against the updated ScrapingToolkit as it changes, support multiple concurrent files, but none of that needs to slow down moving forward on this.

PR #305, opened 2021-02-17
Scraping tool kit

russell-d-e:
Adding the ScrapingToolKit to the master branch


glguy (approved):
I think there are some opportunities for adding features to this, but it's quite useful as it is and I'd like to be able to merge it in so we can build on it more effectively in the ada tools

PR #312, opened 2021-03-01
Add initial CHANGELOG.md draft

tuxji:
Will need review by team members.

In .chglog/CHANGELOG.md, save output from running `git-chglog
--next-tag v5.0`.

In .chglog/CHANGELOG.tpl.md, save template changes to slightly improve
output.

In .chglog/README.md, document how to install and run git-chglog even
though it wasn't that useful after all.

In CHANGELOG.md, save initial draft using only version numbers and
merge commit urls from `.chglog/CHANGELOG.md` and looking at pull
requests' descriptions and diffs to write actual changelog messages.


tuxji:
Please review the CHANGELOG.md in the root directory, not .chglog/CHANGELOG.md.  The latter was automatically generated and used only as an initial draft which I copied to the root directory and edited into its final form.

PR #322, opened 2021-03-16
Updated for MODEL.sadl and refined query

AbhaMoitra:
This updates a SADL data query to use "title" instead of "name" in column header (the CSV file had already been manually updated for RACK V5). Because of new file MODEL.sadl, updated GeneratePropInfoCSV.sadl and re-ran it.


kityansiu (approved):
Looks good!

PR #327, opened 2021-03-22
Assist tool updates

kquick:
* With change of `COMPILE` `compiledBy` to a `wasAssociatedWith`, the range became a TOOL (subclass of AGENT) and needed a way to link back to the `FILE`.  This was done via the `definedIn`, broadening the domain of the latter a bit.
* Update ASSIST analyze and check to read from a  live database (e.g. Fuseki) as well as local OWL files.
* Fix a couple of small issues in the Turnstile ASSIST definitions
* Better documentation for ASSIST tools.
* Fixed ASSIST for the recent SADL change which links defined classes to the file in which they are defined (via an RDF blank node).

### DIFF
--- a/RACK-Ontology/OwlModels/FILE.owl
+++ b/RACK-Ontology/OwlModels/FILE.owl
@@ -96,9 +96,9 @@
     <rdfs:range rdf:resource="#HASH_TYPE"/>
   </owl:ObjectProperty>
   <owl:ObjectProperty rdf:ID="definedIn">
-    <rdfs:comment xml:lang="en">The file an ENTITY is defined or contained in</rdfs:comment>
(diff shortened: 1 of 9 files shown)


glguy (changes requested):
Properties defined with `has a single value` are being checked to have cardinality `1`, but they should have cardinality `0` or `1`.

Our dataset relies on this and it's generating a lot of errors when I run check against our triple-store.

https://www.w3.org/TR/owl2-syntax/#Functional_Data_Properties

glguy (approved):
assist was checking the properties correctly. The issue was in how SADL is generating constraints.

russell-d-e (approved):

PR #346, opened 2021-04-16
some single values should be at most 1 values

glguy:
While the standalone statement `prop has a single value.` on its own means a
property is functional (has 0 or 1 values), the statement
`described with a single value of type` means exactly 1.

This commit adapts some of the instances where 0 or 1 was intended
rather than a strict exactly one. This is a generalization. All previously
valid data will stay valid.


glguy:
(This issue was noticed by running ASSIST's check tool on the default v5.1 loaded dataset where we've got some HAZARDs loaded that do not satisfy their constraints)

kquick (approved):
These look like appropriate changes in the SADL for those elements, as confirmed by the emitted OWL.

AbhaMoitra (approved):

PR #347, opened 2021-04-19
DesCert ontology

bhatt111:
First draft of DesCert ontology overlay.

### DIFF
--- a/DesCert-Ontology/ontology/DesCert.sadl
+++ b/DesCert-Ontology/ontology/DesCert.sadl
@@ -43,11 +43,13 @@ import "http://arcos.rack/MODEL".
        //   identifier
        //   title - name of tool
        //   toolVersion
+       // NOTE: actedOnBehalfOf property of AGENT is not used 
        // We are adding more properties to TOOL here
-       toolInstallationConfiguration describes TOOL with a single value of type string.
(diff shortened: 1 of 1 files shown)


glguy:
This PR is using many cases of `describes with a single value of type` which means there must be exactly 1 of that value.

This is in (unfortunate) contrast to a top-level statement that `someProperty has a single value.` which means 0 or 1, optional.

We very recently fixed a couple of instances of this situation in #346 .

You might change some of these to:

```
prop describes Class with values of type Type.
prop describes Class with at most 1 value.
```

bhatt111:
Hi Eric,
I made changes per your suggestion. Thanks. One additional question I have: I am defining new properties like _sallyUserGuide_, _radlUserGuide_, etc. I am thinking of using a single property name _userGuide_ for these. Is the is the right thing to do?
Devesh

bhatt111:
I made some properties common across multiple classes: userGuide, semantics. More clean up of "single value" usage.

bhatt111:
Added Daikon ontology to DesCertOntology.

bhatt111:
The push 29563e5 just went through when I re-pushed it. This is for adding the Daikon ontology to DesCert.sadl.

PR #357, opened 2021-04-26
query: added Boeing data validation query nodegroup

kityansiu:
Added query nodegroups for doing validation on Boeing data.


AbhaMoitra (approved):

kquick:
I believe that most, if not all of these validations should be handled by the `check` tool.

PR #359, opened 2021-04-27
setup-rack.sh to load LMCO & SRI ontologies; load only DO-178C

kityansiu:


cli/setup-rack.sh
```diff
@@ -24,14 +24,16 @@ rack model import --clear ../RACK-Ontology/OwlModels/import.yaml
...
 
-rack data import --clear ../RACK-Ontology/OwlModels/ARP-4754A.yaml
-rack data import --clear ../RACK-Ontology/OwlModels/DO-330.yaml
+# rack data import --clear ../RACK-Ontology/OwlModels/ARP-4754A.yaml
+# rack data import --clear ../RACK-Ontology/OwlModels/DO-330.yaml
 rack data import --clear ../RACK-Ontology/OwlModels/DO-178C.yaml
-rack data import --clear ../RACK-Ontology/OwlModels/MIL-STD-881D.yaml
-rack data import --clear ../Turnstile-Ontology/99-Utils/Data/Model.yaml
+# rack data import --clear ../RACK-Ontology/OwlModels/MIL-STD-881D.yaml
+rack data import ../Turnstile-Ontology/99-Utils/Data/Model.yaml
```
glguy, line 39: The reason there was a `--clear` on each of these is that each loads into a separate data graph (and the clear only affects that one data graph). I think the clear flag should remain in place.

kityansiu: Ah! Makes sense. I'll add --clear back in.

glguy: Given that this isn't obvious, could you add a comment along the way telling people? Otherwise all these `--clear` are... clearly confusing :)

glguy (approved):

PR #361, opened 2021-04-27
migration: add commit 0a89f70f and freeform notes

Ptival:


glguy (approved):
An escape hatch seems OK

PR #363, opened 2021-04-27
migration: make the output Markdown-ready

Ptival:
This makes it so that we can directly copy and paste the output of the
`rack_migrate --list-changes-only` command into the wiki.


kityansiu:
Just browsed the wiki page and have the following requests:

- Can you state that this is for “core ontology” (and does not cover changes within the ontology overlays)?
- Can the section headings have the format "from version -> to version"? For example, 5.1 -> 6.0. I saw you had that previously. Maybe something's changed.

tuxji:
It would be nice if you could run [markdownlint](https://github.com/DavidAnson/markdownlint) on the generated wiki page and make the necessary changes to the `rack_migrate --list-changes-only` command to fix or disable markdownlint's warnings.

For example, you could ensure the output begins with these lines to disable some of markdownlint's warnings:

```text
<!-- markdownlint-disable first-line-heading -->
<!-- markdownlint-disable line-length -->
<!-- markdownlint-disable no-bare-urls -->
```

Then you could fix the remaining warnings by ensuring:

1. There is a blank line between the `Commit` lines and the lines beginning with dashes (the commit's changes).
2. The output ends with a single newline (not a text line without a newline).

That would shut up all of the warnings.

tuxji:
Thank you, now <https://github.com/ge-high-assurance/RACK/wiki/RACK-ontology-detailed-changelogs> has no markdownlint warnings!  I'm preparing the RACK v6.0 release so I looked at the changelog and I don't notice any v5.1 -> v5.9 entries. That's a time period two months long - shouldn't there be some v5.1 -> v5.9 entries there?

tuxji:
Actually, I looked at the last [entry](https://github.com/ge-high-assurance/RACK/commit/0a89f70ff929380269a79fe2fc82f5dde346ed8c) in the changelog and it was merged into the repository 10 days ago.  I think the version headings may simply be positioned incorrectly.  Should the v5.1 -> v5.9 heading be where the v5.0 -> v5.1 heading is?

Ptival:
Initially, I had meant for these lines to not be headings, but mark the point in the document at which we cross a tag. So they were not supposed to mean "under this heading, changes for tag v5.9" but rather "upon reaching this marker, we cross tag v5.9".

I see how this is extremely ambiguous with how headings are generally used. So instead I could make it so the heading is very explicit, something like "Changes between version X and Y", in which case I do need to offset them.

Ptival:
How about the current formatting, where most recent changes are more prominently displayed?

https://github.com/ge-high-assurance/RACK/wiki/RACK-ontology-detailed-changelogs

tuxji:
Yes, I see how you originally had only one tag per heading, then you added a second tag at Kit's suggestion.  I agree with the change you made to go back to one tag per heading, saying "Changes for <tag>", displaying most recent changes first.  That is unambiguous too.  Thanks!

glguy (approved):

PR #369, opened 2021-05-03
changed casing in SACM ontology overlay

kityansiu:
I gave the LMCO team wrong advice about capitalization on Classes. All caps are only used on classes in the core ontology. This corrects the casing. Sorry about that!

### DIFF
--- a/LM-Ontology/ontologies/SACM.sadl
+++ b/LM-Ontology/ontologies/SACM.sadl
@@ -6,186 +6,186 @@ import "http://arcos.rack/PROV-S".
 
 /* Base Classes. */
 
-ELEMENT is a type of ENTITY.
-        uuid describes ELEMENT with a single value of type string.
+Element is a type of ENTITY.
+        uuid describes Element with a single value of type string.
 
-MULTI_LANG_STRING
-    is a type of ELEMENT.
-        hasValue describes MULTI_LANG_STRING with values of type LANG_STRING.
+Multi_Lang_String
+    is a type of Element.
+        hasValue describes Multi_Lang_String with values of type Lang_String.
 
-LANG_STRING
-    is a type of ELEMENT.
-        lang describes LANG_STRING with values of type string.
-        lang of LANG_STRING has at most 1 value.
-        contents describes LANG_STRING with values of type string.
-        contents of LANG_STRING has at most 1 value.
+Lang_String
+    is a type of Element.
+        lang describes Lang_String with values of type string.
+        lang of Lang_String has at most 1 value.
+        contents describes Lang_String with values of type string.
+        contents of Lang_String has at most 1 value.
 
-EXPRESSION_LANG_STRING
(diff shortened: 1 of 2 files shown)


LM-Ontology/ontologies/SACM.sadl
```diff
@@ -6,186 +6,186 @@ import "http://arcos.rack/PROV-S".
...
+Artifact
+    is a type of Artifact_Asset.
+        artifactVersion describes Artifact with values of type string.
+        artifactDate describes Artifact with values of type string.
 
-SACM_ACTIVITY
-    is a type of ARTIFACT_ASSET.
-        startTime describes SACM_ACTIVITY with values of type string.
-        endTime describes SACM_ACTIVITY with values of type string.
+Activity
```
glguy, line 175: Was this supposed to lose its `SACM_` prefix? Some stuff has the prefix and some doesn't; I don't understand the pattern.

LM-Ontology/ontologies/SACM.sadl
```diff
@@ -6,186 +6,186 @@ import "http://arcos.rack/PROV-S".
...
-        hasDescription describes MODEL_ELEMENT with values of type DESCRIPTION.
-        hasNote describes MODEL_ELEMENT with values of type NOTE.
-        hasName describes MODEL_ELEMENT with values of type LANG_STRING.
-
-ARTIFACT_ELEMENT
-    is a type of MODEL_ELEMENT.
+Element is a type of ENTITY.
+        uuid describes Element with a single value of type string.
+
+Multi_Lang_String
```
glguy, line 12: I think the Turnstile example would have used `MultiLangString` instead of the underscored version. Is that divergence intended?

glguy (approved):
Assuming the answers to my two questions are *no change needed* this PR seems fine.

LM-Ontology/ontologies/SACM.sadl
kityansiu: I don't know about pattern. But you were right about the missing `SACM_` prefix on this one. It was my copy-paste error. Good catch!

LM-Ontology/ontologies/SACM.sadl
kityansiu: Another good catch! I'll make the mods.

carter-e-veldhuizen (approved):
Changes look good to me, thanks for tagging my review so I could see this. When should we expect this to make it into RACK?

kityansiu:
We will merge this into the dev branch. It will be part of RACK v7.0 with a targeted release date of 6/7.

tuxji:
FYI, every time we merge a pull request into the dev branch, our CI workflow pushes a new Docker image with the tag gehighassurance/rack-box:dev to Docker Hub.  Just letting you know in case you don't want to wait for RACK v7.0 on 6/7.

PR #390, opened 2021-05-13
assist: fix exports in the Prolog ontology generation

Ptival:
The original code was producing incorrect results when there were no
properties, as the last thing would have a trailing comma, not followed by any
property.

This improved version first aggregates together all things and properties
strings, then uses a combinator to insert commas as appropriate.

For an example of where this went wrong, try and generate software.pl with the
current state of the ontology.

### DIFF
--- a/assist/bin/rack/write_ontology.pl
+++ b/assist/bin/rack/write_ontology.pl
@@ -9,10 +9,12 @@
 % material are those of the author(s) and do not necessarily reflect the views
 % of the Defense Advanced Research Projects Agency (DARPA).
 
-:- module(write_ontology,
-          [
-           write_ontology/0
-          ]).
+:- module(
+    write_ontology,
+    [
+        write_ontology/0
+    ]
+).
 
 :- ensure_loaded('../paths').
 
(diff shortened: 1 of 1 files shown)


assist/bin/rack/write_ontology.pl
```diff
@@ -153,12 +161,12 @@
...
     writeln(Handle, '% THIS FILE WAS AUTOMATICALLY GENERATED, SEE README'),
     nl(Handle),
-    atomic_list_concat([':- module(', Module, ','], FirstLine),
-    writeln(Handle, FirstLine),
-    writeln(Handle, '          ['),
-    write_exports_for_things(Handle, Things),
-    write_exports_for_properties(Handle, Properties),
-    writeln(Handle, '          ]).'),
+    writeln(Handle, ':- module('),
+    atomic_list_concat(['    ', Module, ','], ModuleLine),
```
kquick, line 165: The indentation here is still hard-coded, relative to what `export_indentation` will provide; it's not terribly important, but at some point we might want to pass an indentiation level that gets incremented and decremented.  Just a comment though, probably not worth spending time on right now.

kquick (approved):

assist/bin/rack/write_ontology.pl
Ptival: That's actually fairly easy to do in a nicer way, fixed it in 8b9726f

kquick: Nice!

PR #394, opened 2021-05-17
setup-rack.sh loads core ontology only

kityansiu:
This PR does the following:
- cli/setup-rack.sh loads only the core ontology
- cli/setup-arcos.sh loads the ARCOS ontology overlays
- Turnstile ontology + data is now done with Turnstile-Ontology/99-Utils/Load-TurnstileData.sh

Other things that had to be reorg'ed to make this work:
- nodegroups directory now has subdirectories to hold auto generated nodegroups from the overlays, separate from the auto generated nodegroups from the core <-- @cuddihyge, please confirm that this structure will work for you for future auto CDR generation
- OwlModels files in each of the ARCOS overlays are cleaned up. (Only 2 files: import.yaml and corresponding overlay .owl)
- import.yaml in each of the overlay directories had to be updated because this file will now reside in OwlModels in each of the overlay directories 

Hope each of the reviewers can take a good look and try out the changes to make sure that nothing is misplaced/missing.


cuddihyge:
@kityansiu Regarding the CDR and nodegroup sub-directories, we'll have to figure out a maintainable way to map the base URI's of the ontology classes into the right sub-directory.   It would be simple to one-off a hardcoded solution, but it would be less work in the long run to do some straight-forward transformation of the class URI into a sub-directory name.

For example, we could use the "USE_THIS" portion of http://USE_THIS/CLASSNAME
and we'd end up with sub-directories:
- arcos.rack
- arcos.AH-64D
- arcos.arbiter
- arcos.descert
- Turnstile

or something simple-to-automate  along these lines.

cli/setup-arcos.sh
```diff
@@ -0,0 +1,47 @@
...
+rack nodegroups import ../nodegroups/ingestion/SRI
+
+# STR model & auto-generated nodegroups
+rack model import ../STR-Ontology/OwlModels/import.yaml
+rack nodegroups import ../nodegroups/ingestion/STR
+
+# load applicable standards
+# rack data import --clear ../RACK-Ontology/OwlModels/ARP-4754A.yaml     # from datagraph http://rack001/arp-475a
+# rack data import --clear ../RACK-Ontology/OwlModels/DO-330.yaml        # from datagraph http://rack001/do-330
+rack data import --clear ../RACK-Ontology/OwlModels/DO-178C.yaml         # from datagraph http://rack001/do-178c
```
glguy, line 46: Should this be loaded by the Turnstile script, instead?

glguy (approved):

cli/setup-arcos.sh
kityansiu: That's a good question. @russell-d-e, when do you think DO-178C should be loaded? When setting up ARCOS or when loading Turnsile data?

russell-d-e: I think it should be loading by whomever is using.  So it should be loaded by Turnstile and if someone is going to related there generated evidence then should include it as well.  If there is multiple providers wanting to use it they all should include it as there ingestion package.  As long as everyone uses the same data there should not be any duplicates (as long as the nodegroups are constructed with a lookup).

PR #398, opened 2021-05-20
SRI ontology for review

bhatt111:
This is the ontology update based upon comments from the last review with TA2.


AbhaMoitra:
Couple of items that need to be changed. 
* In line 389, the syntax should be "{SourceCode or RandoopTestsAndMetrics}" 
* The property "verifies" is defined in the core ontology and this ontology is changing the domain of this property. The 2 domains for "verifies" in this ontology are DaikonInvariantOutput and RandoopTestsAndMetrics. Both of these are types of ENTITY. Either they should both be of type TEST or use a qualifier on "verifies" or rename "verifies".

bhatt111:
Made fixes for comments made by Abha.

AbhaMoitra:
@glguy : I have looked over the ontology and it seems fine. However one of the CI workflow action is failing - can you take a look at it?

glguy:
> @glguy : I have looked over the ontology and it seems fine. However one of the CI workflow action is failing - can you take a look at it?

This is because the PR is on an external branch and doesn't have access to the secret token our CI uses. I agree that the change seems fine; I wouldn't necessarily block it on CI. If you want to be sure we can make a local copy of the branch to run CI on.

tuxji:
@AbhaMoitra , the CI workflow failed because the pull request was from a forked repository.  We want to support running CI on pull requests from forked repositories.  I believe I've fixed the problem in PR #401.  @bhatt111 , please merge our repo's master branch into your repo's master branch and push again to see if #401 lets your pull request's next CI job run successfully.

PR #404, opened 2021-05-21
Changed and moved definedIn property.

AbhaMoitra:
As we discussed in the team meeting, I have changed and moved the definedIn property.


glguy:
This changes the domain of definedIn from `THING` to `ENTITY`. This could preclude some cases like a software agent defined in a file. Is this intentional?

AbhaMoitra:
@glguy : Yes it does change the domain. But this is what was decided at the Tuesday meeting. At least that is what is in the meeting notes. Do we want to review it with the team on 5/25? The "software agent defined in a file" would still be allowed as FILE is a subClass of ENTITY.

glguy (approved):

PR #408, opened 2021-05-25
Implements changes for issue 383

russell-d-e:
Updated sadl to have Wg-63 acting on behalf of S-18 and removed S-18 as an issuingOrganization

### DIFF
--- a/RACK-Ontology/OwlModels/ARP-4754A.owl
+++ b/RACK-Ontology/OwlModels/ARP-4754A.owl
@@ -239,10 +239,11 @@
       <j.0:AGENT rdf:ID="SAE"/>
     </D:approvalAuthority>
     <D:issuingOrganization>
-      <j.0:AGENT rdf:ID="WG-63"/>
-    </D:issuingOrganization>
-    <D:issuingOrganization>
-      <j.0:AGENT rdf:ID="S-18"/>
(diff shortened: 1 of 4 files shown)


AbhaMoitra (commented):
The changes look fine. But it failed a check - need to check that further?

tuxji:
I took a quick look.  The lint was checking only shell scripts and nothing in this PR could have affected the shell scripts.  This PR ran the continuous workflow twice and the lint check succeeded the second time, so I think the lint failure was a fluke.

kquick:
The lint failure is occurring on the `push` build branch, but not the `pull` branch.  The difference here is that the `push` branch builds the exact commit pushed, but the `pull` branch builds the speculative merge of the pull with `master`.  This particular PR is based on a fairly old point in `master` (https://github.com/ge-high-assurance/RACK/commit/55edd52510bdfc5049a45a88fec717821a195ac3) and does not include https://github.com/ge-high-assurance/RACK/pull/348, which provides the fix for shellcheck updating from version 0.7.1 to 0.7.2.

It's worth noting that although we've locked the version of the shellcheck github action to 1.0.0, the action downloads the latest release of shellcheck itself and there doesn't appear to be any way to control this.

tl;dr ignore the failing check: it will be resolved on merging to master.

kquick (approved):

PR #419, opened 2021-05-28
migration: show commit subject

Ptival:
These are the changes that generated the new format for the https://github.com/ge-high-assurance/RACK/wiki/RACK-ontology-detailed-changelogs page.

### DIFF
--- a/migration/rack_migrate/rack_migrate
+++ b/migration/rack_migrate/rack_migrate
@@ -20,7 +20,8 @@ from typing import Callable, List, NoReturn, Tuple
 import semtk
 
 import git_helpers as git
-from ontology_changes.ontology_change import Commit, logger, stylize_file_name
+from ontology_changes.ontology_change import Commit, LOGGER_ID, logger, stylize_file_name
+from rack_crawl.crawler import commit_touches_ontology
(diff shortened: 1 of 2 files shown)


migration/rack_migrate/rack_migrate
```diff
@@ -160,6 +161,13 @@ if len(commits_to_consider_in_chronological_order) == 0:
     sys.exit(1)
 
 
+def markdown_link_to_commit(
+    commit_id: str,
+    link_text: str,
+) -> str:
+    return f"[{link_text}](https://github.com/ge-high-assurance/RACK/commit/{commit_id})"
```
glguy, line 168: Is it possible for `link_text` to contain a `]`?

glguy (approved):
The PR is good assuming the question above doesn't require a fix.

migration/rack_migrate/rack_migrate
Ptival: I think it might, good catch!

glguy:
@Ptival I updated this branch to escape `[` and `]`, which seems to work well enough in the GitHub renderer.

PR #420, opened 2021-05-28
update cli/README.md

Ptival:
The README contained obsolete folder structures.

As an aside, I always find it confusing that the **data** import script for Turnstile is called `Model.yaml`, while the **model** import script is called `import.yaml`. Not sure how the current naming came about, but I'd vote to change it.


glguy (approved):
Good catch.

PR #421, opened 2021-05-31
migration: update README

Ptival:
Mostly because the note at the top was outdated, but I also changed a couple things.

### DIFF
--- a/migration/README.md
+++ b/migration/README.md
@@ -14,25 +14,24 @@ help.
 ## How to build?
 
 To set up, we suggest using Python's virtualenv.  A possible setup is to run
-(replacing `<env>` with the name of the folder you want your virtual environment
+(replacing `env` with the name of the folder you want your virtual environment
 saved in):
 
 ```
-python3 -m venv <env>       # Creates a virtual environment in the 'env' directory
-source <env>/bin/activate   # Sets up the virtual environment in the **current** shell
+python3 -m venv env       # Creates a virtual environment in the 'env' directory
+source env/bin/activate   # Sets up the virtual environment in the **current** shell
 ```
 
-The rest of the procedure uses `pip`.  You can choose to install it globally
-using your package manager of choice, or locally via `python3 -m pip install
-pip`, or you can just always use `python3 -m pip` where we use `pip` in the
-following commands.
+The rest of the procedure uses `pip`, which is added to your `PATH` by the
+virtual environment activation command.  If you open a new shell, make sure to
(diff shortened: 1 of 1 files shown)


migration/README.md
```diff
@@ -1,34 +1,44 @@
...
 
-To set up, we suggest using Python's virtualenv.  A possible setup is to run:
+To set up, we suggest using Python's virtualenv.  A possible setup is to run
+(replacing `<env>` with the name of the folder you want your virtual environment
+saved in):
 
 ```
-python3 -m venv env       # Creates a virtual environment in the 'env' directory
-source env/bin/activate   # Sets up the virtual environment in the **current** shell
+python3 -m venv <env>       # Creates a virtual environment in the 'env' directory
```
glguy, line 21: I think it makes sense to be general here, but that stops people from copy/pasting as easily and it seems quite rare that people will need multiple virtual environments. I'm OK to leave it like this if you prefer it though.

migration/README.md
```diff
@@ -1,34 +1,44 @@
...
+source <env>/bin/activate   # Sets up the virtual environment in the **current** shell
 ```
 
+The rest of the procedure uses `pip`.  You can choose to install it globally
```
glguy, line 25: Once we're in a venv doesn't `pip3` come from the venv? Is there any point to using the `python3 -m pip` invocation?

migration/README.md
```diff
@@ -1,34 +1,44 @@
...
+
 In a shell where the environment is activated, you ought to be able to:
 
 ```
-python3 -m pip install -r requirements.txt
-# or, if you intend to modify the tool:
-python3 -m pip install -r requirements.txt -r requirements-dev.txt
+pip install -r requirements.txt
+# or, if you intend to work on the tool:
+pip install -r requirements.txt -r requirements-dev.txt
```
glguy, line 35: For working on the tool we might even recommend installing it with `-e`

glguy (approved):

migration/README.md
Ptival: That makes sense, I think the comment makes it explicit enough that `env` is something under your control.

migration/README.md
Ptival: I did not know about `-e`, sounds neat!

migration/README.md
Ptival: Indeed that's the case, I did not realize, will simplify explanations!

PR #430, opened 2021-06-02
Update ASSIST runtime help and allow skip of datadir.

kquick:


glguy (approved):
Works for me with and without `-d` and the new default makes more sense to me

PR #432, opened 2021-06-02
assist: write_ontology now exports RDF assertions

Ptival:
The previous write_ontology exported what I refer to as RDF queries for each
class and property, that is, a Prolog predicate that allows one to retrieve
instances of a class of property that are registered within the database.

This adds a notion of RDF assertions for each class and property, allowing one
to register new data in the database.  As usual, this was already doable via
the ASSIST tooling, this commit just adds convenience operators for
RACK-specific data.

### DIFF
--- a/assist/bin/rack/write_ontology.pl
+++ b/assist/bin/rack/write_ontology.pl
@@ -86,19 +86,35 @@ ontology_file_path(Namespace, FilePath) :-
     directory_file_path(OntologyDir, FileName, FilePath).
 
 
-%! export_thing(+Thing, -Export) is det.
+%! export_class(+Class, -Export) is det.
 %
-%    Gives the export string for a thing.
-export_thing(Thing, Export) :-
-    string_lower(Thing, LowercaseThing),
-    atomic_list_concat([LowercaseThing, '/1'], Export).
+%    Gives the export string for RDF query of a class.
+export_class(Class, Export) :-
+    string_lower(Class, LowercaseClass),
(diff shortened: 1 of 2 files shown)


assist/bin/rack/write_ontology.pl
```diff
@@ -84,21 +84,37 @@
...
 
 
-%! export_thing(+Thing, -Export) is det.
+%! export_class(+Class, -Export) is det.
 %
-%    Gives the export string for a thing.
-export_thing(Thing, Export) :-
-    string_lower(Thing, LowercaseThing),
-    atomic_list_concat([LowercaseThing, '/1'], Export).
+%    Gives the export string for RDF query of a class.
```
kquick, line 89: For export_class and export_property, you now have assert forms and the docs indicate the old are queries and the new are assertions, but you don't really say what an assertion is.  You might want to indicate that a query succeeds only if the RDF triple already exists, and the assert creates a new RDF triple (or maybe provide a reference to rdf/3 and rdf_assert/3.

kquick (approved):
Just a suggestion about updating some of the comments.

assist/bin/rack/write_ontology.pl
Ptival: I like your suggestion, though I felt like this information would be most useful written in the created files, so that people seeing `assert_file` understand what it attempts to do.

I've pushed a commit that adds such comments in the generated files. What do you think?

kquick: Looks good!

PR #433, opened 2021-06-02
assist: fix write_ontology handling of dashes in modules

Ptival:
Prolog modules must declare their name as a Prolog atom.  Unfortunately, dashes
in atoms have a special meaning, and so atoms whose name must contain dashes
have to be properly quoted.

write_ontology was generating such a module named prov-s for PROV-S.owl, and so
the generated Prolog file was incorrect.  This changes makes sure that we add
quotes around such names when they contain dashes, thus generating 'prov-s'
which is a correct atom.

### DIFF
--- a/assist/bin/rack/write_ontology.pl
+++ b/assist/bin/rack/write_ontology.pl
@@ -168,6 +168,16 @@ write_property(Handle, Namespace, Property) :-
     property_declaration(Namespace, Property, Declaration),
     writeln(Handle, Declaration).
 
+
+%! use_quotes_if_contains_dashes(+In, -Out) is det.
+%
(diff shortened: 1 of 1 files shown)


assist/bin/rack/write_ontology.pl
```diff
@@ -168,14 +168,30 @@
...
     ontology_file_path(Namespace, FilePath),
     open(FilePath, write, Handle),
-    string_lower(Namespace, Module),
+    % trick for "function composition"
+    foldl(
+        call,
+        [string_lower, use_quotes_if_contains_dashes],
+        Namespace,
+        Module
+    ),
```
kquick, line 194: Oh, I like this form!

kquick (approved):

Issue #424, opened 2021-06-01
Running ASSIST-DV against RACK 6.0 shows errors


kquick:
Startup the docker image:

```
$ docker pull gehighassurance/rack-box:v6.0
$ docker run --detach -p 80:80 -p 12050-12092:12050-12092 -p 3030:3030 gehighassurance/rack-box:v6.0
$ assist/bin/check -v -m http://localhost:3030/
```

This returns 41 rack definitions lacking a Note/Description and 453 entries with property issues (e.g. `ERROR: http://semtk.research.ge.com/generated#semtk_ng_3ad47533-377a-4877-8699-adfbaee7164d . http://research.ge.com/semtk/prefabNodeGroup#stringChunk has 0 values but cardinality of 1`)

I'm not sure what is generating the above data, but we should probably ensure it doesn't occur in Release 7.

kquick:
These same errors appear in v6.9.  We have some data instances that don't look like they belong in RACK:

  * 2 in http://com.ge.research/sample/hardware#
  * 3 in http://com.ge.research/sample/testconfig#
  * 91 in http://demo/prefix#
  * 3 in http://research.ge.com/semtk/services/job#
  * 242 in http://semtk.research.ge.com/generated#

The two in `hardware`:
```
http://com.ge.research/sample/hardware#complete rdf:type http://com.ge.research/sample/hardware#HardwareModStatus
http://com.ge.research/sample/hardware#inProgress rdf:type http://com.ge.research/sample/hardware#HardwareModStatus
```

The three in `http://com.ge.research/sample/testconfig#` are all instances of `http://com.ge.research/sample/testconfig#TableType` and the instances are: `#Average`, `#HourMax`, and `#Interpolated`

The three in semtk are all instances of `http://research.ge.com/semtk/services/job#Status` and the instances are: `#Failure`, `#InProgress`, and `#Success`.

@cuddihyge , does SemTK store nodegroup information in the triplestore (wondering about the 242 instances in the `http://semtk.research.ge.com/generated#` namespace)?

PR #436, opened 2021-06-03
[ASSIST-DV] Skip various background data instances.

kquick:
These seem to be generated from SemTK and/or SADL builtins.
This is a workaround for issue #424.


glguy (approved):
Works for me!

PR #450, opened 2021-06-24
Scraping tool kit improvements

russell-d-e:
Improvements to the scraping tool kit, including:
--Auto-generation now works off looking for nodegroups in the nodegroups directory of the RACK repo to generate the Add.py
--Ingestion activities are now created automatically as to produce the dataInsertedBy infromation


kityansiu (approved):
Just tried out ReadNodegroups.py with a new set of turnstile nodegroups. Works great! Though I was thinking it will still be good to keep GenFiles.py (or something equivalent) that uses the data dictionary (table.csv) downloaded from SemTK, cause I found that useful to work with while I was waiting for the new set of CDRs and nodegroups.

Issue #472, opened 2021-07-01
ASSIST - Prolog documentation


Robert-Adelard:
If I run the Prolog documentation server, as per the instructions here

https://github.com/ge-high-assurance/RACK/tree/master/assist/bin#prolog-api-usage

I get a series of errors as follows:

```
ERROR:  .../RACK/assist/bin/documentation/serve.pl:20:
ERROR:    import/1: No permission to import requirements:mitigates/2 into user (already imported from hazard)
ERROR: .../RACK/assist/bin/documentation/serve.pl:20:
ERROR:    import/1: No permission to import requirements:assert_author/2 into user (already imported from hazard)
ERROR: .../RACK/assist/bin/documentation/serve.pl:20:
ERROR:    import/1: No permission to import requirements:assert_mitigates/2 into user (already imported from hazard)
ERROR: .../RACK/assist/bin/documentation/serve.pl:20:
ERROR:    import/1: No permission to import requirements:author/2 into user (already imported from hazard)
ERROR: .../RACK/assist/bin/queries/hazard_structure.pl:18:
ERROR:    import/1: No permission to import requirements:mitigates/2 into hazard_structure (already imported from hazard)
ERROR: .../RACK/assist/bin/queries/hazard_structure.pl:18:
ERROR:    import/1: No permission to import requirements:assert_author/2 into hazard_structure (already imported from hazard)
ERROR: .../RACK/assist/bin/queries/hazard_structure.pl:18:
ERROR:    import/1: No permission to import requirements:assert_mitigates/2 into hazard_structure (already imported from hazard)
ERROR: .../Work/RACK/assist/bin/queries/hazard_structure.pl:18:
ERROR:    import/1: No permission to import requirements:author/2 into hazard_structure (already imported from hazard)
```

Robert-Adelard:
Incidentally, since the Prolog ontology files are no longer included in the RACK distribution, as per

https://github.com/ge-high-assurance/RACK/commit/b90f96c4df07c8c72ab7b5b3c47e4cbd2e96f526

the instructions for running the documentation server should mention the need to auto-generate the ontology files.

Ptival:
Indeed that's embarassing, hah Prolog!

I will see if I can conditionally list the ontology files based on whether they have been generated or not.

Will update.

Robert-Adelard:
Thanks for looking into this - I ran into another problem today that might be related:

```
?- [assist/bin/ontology/prov-s].
[...]

PR #476, opened 2021-07-01
prolog: special case renaming the system module into rack-system

Ptival:
SWI-Prolog refuses to load a file, using 'load_files', that declares a module
named 'system'.  This makes it so the 'system' module is renamed 'rack-system'.

The generated file is still named 'system.pl' and loaded via
'ontology(system)', so this change should be transparent for most uses.


Ptival:
Together with #475 , this should fix the Prolog documentation and #472

kquick (approved):

PR #521, opened 2021-07-26
data_shape: use setup.cfg

Ptival:
I had trouble setting up the environment with the given setup.py.  I went to
read the documentation for setuptools, found out they recommended using
setup.cfg nowadays, switched the configuration for testing purposes, and it
worked.

It's very similar but makes the configuration a proper data file rather than
some constants in a piece of code.  If everyone is happy switching to
that, here is how we could go about it.


glguy (approved):
works for me and I'm quite happy to update to the new practice

PR #539, opened 2021-08-09
migration: relax required constraints in SemTK JSON schema

### LINKED ISSUE #538: SemTKJSON too restrictive
A user of the migration tool has reported the following errors:

```
sNodeGroup -> sNodeList -> 0 -> propList -> 0 -> fullURIName
  field required (type=value_error.missing)

sNodeGroup -> sNodeList -> 1 -> propList -> 0 -> fullURIName
  field required (type=value_error.missing)

sNodeGroup -> sNodeList -> 3 -> propList -> 0 -> fullURIName
  field required (type=value_error.missing)

sNodeGroup -> sNodeList -> 4 -> propList -> 1 -> fullURIName
  field required (type=value_error.missing)

sNodeGroup -> sNodeList -> 4 -> propList -> 2 -> fullURIName
  field required (type=value_error.missing)

sNodeGroup -> sNodeList -> 4 -> propList -> 3 -> fullURIName
  field required (type=value_error.missing)

importSpec -> nodes -> 0 -> props -> 0 -> mapping -> 0 -> colId
  field required (type=value_error.missing)

importSpec -> nodes -> 2 -> props -> 0 -> mapping -> 0 -> colId
  field required (type=value_error.missing)
```

I will update the JSON schema to be more lax about those fields.

Ptival:
Should fix #538

@cuddihyge do these seem reasonable?

I was a bit surprised by missing `colId` on `mappings`.

### DIFF
--- a/migration/ontology_changes/change_property_range.py
+++ b/migration/ontology_changes/change_property_range.py
@@ -104,6 +104,7 @@ class MigrationVisitor(semtk.DefaultSemTKVisitor):
 
             # cleanup
             for sparqlID in node.SnodeSparqlIDs:
+                if json.importSpec is None: continue
                 for index, importSpecNode in enumerate(json.importSpec.nodes):
                     if importSpecNode.sparqlID == sparqlID:
                         importSpecNode.type = self.rename(
--- a/migration/ontology_changes/delete_property.py
+++ b/migration/ontology_changes/delete_property.py
@@ -61,6 +61,7 @@ class MigrationVisitor(semtk.DefaultSemTKVisitor):
 
                 # cleanup
                 for sparqlID in node.SnodeSparqlIDs:
+                    if json.importSpec is None: continue
                     for index, importSpecNode in enumerate(json.importSpec.nodes):
                         if importSpecNode.sparqlID == sparqlID:
                             field = stylize_json("sparqlID")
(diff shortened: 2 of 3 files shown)


cuddihyge:
I don't think this will hurt anything, but I also believe that tighter checking on the JSON would be helpful, not looser.    There are some places that should check it but do not.
I'd like to understand our what led to this and other options for fixing it.

Ptival:
What led to this was a Boeing person trying to run the migration script, and hitting errors on all of those cases, supposedly because their nodegroup has missing values in those places.

I can certainly ask how they got such a nodegroup, or put them in contact with you if this involves information I don't have access to.

cuddihyge:
I wouldn't mind reviewing a copy of one of those nodegroups along with whatever they know about where it came from.
If they were manually edited, we can stop there.   

But it is possible that some delete functionality in the ingestion template is leaving some garbage.  In that case SemTK might be seeing that a higher level structure is gone and then ignore the garbage, where your validation still catches the inconsistency.    If so, I'd like to fix it.

The SemTK service layer currently does **not** do a formal step of validating the JSON when it arrives.  Perhaps it should.  
1. It was never anticipated that people would manually edit the JSON
2. Performance cost needs to be weighed against the benefits

Ptival:
According to my Boeing contact, the nodegroup is obtained via GrammaTech tooling, and they suggest that Greg Nelson would be the point of contact for more details about the tool.

PR #541, opened 2021-08-18
Remove subtypes of wasGeneratedBy

glguy:
These subtypes are not offering a semantic refinement and are being used as aliases for the underlying `wasGeneratedBy` relationship that should be used in place of these aliases.


kityansiu:
I added this PR to Pre-Release 9 so we don't lose track of it if we want to get it into the next release.

glguy:
GramaTech uses `http://arcos.rack/TESTING#producedBy`, SRI uses `http://arcos.rack/SYSTEM#producedBy`

I think this change is worth doing, but we'll want these teams to update their ingestion process not to use the subrelations.

cuddihyge:
@tuxji John, are these being generated automatically during the CI yet, or should we still build them manually and check them in.   
(this is the IngestTemplateGenerator step that builds CDR's)

tuxji:
Not yet.  I will start making the changes needed to generate the nodegroups automatically.

PR #565, opened 2021-09-30
mutualize CLI checks in RACK scripts

Ptival:


glguy:
Did we already do the change that all the scripts in cli/ get installed in the docker image?

Ptival:
I believe this is a manual step?

The [README](https://github.com/ge-high-assurance/RACK/blob/master/rack-box/README.md) still lists:

`tar cfz RACK/rack-box/files/rack-cli.tar.gz RACK/cli/{setup-rack.sh,wheels}`

But looking at the 7.9 pre-release image, I see 4 shell files:

```
save-nodegroups.sh  setup-arcos.sh  setup-clear.sh  setup-rack.sh
```

We should copy more in this case.

Ptival:
I have now stumbled upon [this README](https://github.com/ge-high-assurance/RACK/blob/master/Turnstile-Example/Turnstile-IngestionPackage/README.md) which made me realize that there is also [this script](https://github.com/ge-high-assurance/RACK/blob/master/GE-Ontology/setup-GE.sh) that lives on its own in the `GE-Ontology` folder and has the same type of code.

PR #567, opened 2021-10-01
setup.cfg for migration

Ptival:
Switching to declarative format for uniformly across code base.


glguy (approved):
Builds for me, and it's good to see more of our stuff using setup.cfg

PR #568, opened 2021-10-01
RACK CLI: Use utf-8-sig

glguy:
This change makes our tools tolerant of the signature Microsoft tools like to add to CSV files without requiring that signature.


Ptival (approved):
Nice!

Issue #578, opened 2021-11-03
add property for filepath in DOCUMENT


kityansiu:
From [Summary-Learnings, Gaps, Enhancements](https://github.com/ge-high-assurance/RACK/wiki/Summary-Learnings,-Gaps,-Enhancements):
> [Adelard] DOCUMENT should have a URL attribute that can be used to identify / access the document

Add a property in DOCUMENT for the filepath (i.e., a place to specify the URL where the document can be identified/accessed). Note: we have something similar in FILE --
`filename (note "A logical file path relative to the container of the file.") describes FILE with a single value of type string.`

PR #593, opened 2021-11-16
Add entityURL property

glguy:
Add a property for external access URL to entities

See #578


kquick:
This also seems like a good way to address the concern in the first full paragraph on page 19 of the phase 1 assessment report.

PR #595, opened 2021-11-22
Turnstile req ingestion

Herencia:
Please review the code


kityansiu:
1) Can you please double check what's being created by the "Low-level Requirements" portion? What I see is there are still HLR-1, HLR-2, and HLR3, but I expect HLR-1 to be replaced with HLR-1:v1 and HLR-1:v2, and HLR-2 replaced by HLR-2:v1 and HLR-2:v2. You can see the presence of HLR-1, -2, -3 in \TurnstileLowLevelRequirements\turnstile_HighLevelRequirement1.csv and \TurnstileLowLevelRequirements\turnstile_HighLevelRequirement2.csv.
2) Can you please add HLR-3:v1 and HLR-3:v2? This is because later on lines 423 and 429 expects an HLR-3:v1, and also in TurnstileRequirementModel.

Herencia:
I changed the Python code to fix the issues of updating  the requirement versions.

Issue #576, opened 2021-10-27
add architecture to ontology


kityansiu:
From [Summary-Learnings, Gaps, Enhancements](https://github.com/ge-high-assurance/RACK/wiki/Summary-Learnings,-Gaps,-Enhancements):

> [SRI] Architecture: System or software architecture specification also deserve first-class status in the ontology. In SRI overlay, we added a class and activities for SystemArchitecture, with specific relationships to entities such as SYSTEM.

Will be in Boeing.sadl in v9.0. But see also Data Model Proposal Board ([card](https://github.com/ge-high-assurance/RACK/projects/8#card-72164022)).

Issue #573, opened 2021-10-27
add TEST_PROCEDURE to the core ontology


kityansiu:
From [Summary-Learnings, Gaps, Enhancements](https://github.com/ge-high-assurance/RACK/wiki/Summary-Learnings,-Gaps,-Enhancements):

> [Adelard] DO-178C distinguishes between a Test case and a Test procedure. Roughly speaking, a Test case is the specification of a test and a Test procedure is an implementation of a test – logically, these are not the same thing.

> [SRI] Test Obligations, Test Cases, and Test Procedures: There is a single class TEST in RACK that is not sufficient to capture the attributes semantics across all the testing-related entities, as others have also noted. It would help to use generic guidance from DO-178C here. In Do-178C, a test case is an abstract specification that consists of test criteria for requirement coverage (i.e., what aspect of requirement clause/subclause does the test cover), test description, trace to the requirement or to test obligation (oracle), test inputs values and expected output values applied to the component under test. A test procedure is typically a script that applies the input values to the component, measures output values and compares them to expected outputs of test case. One may merge test case and procedure as long as all the relevant semantics are captured; but the TEST class in RACK doesn’t have those attributes.

This was discussed at the Data Model Decision Team Meeting on 10/25/21. We will look into including Test Procedure in the core ontology.

AbhaMoitra:
At the team meeting on 11/9/2021 we also decided to ad PARTITION as part of this.

PR #597, opened 2021-11-23
Issue573

russell-d-e:
#573 and #576 

Added ontology for Test Procedure and test Logs.

Also created new Ontology for Hardware Components


AbhaMoitra:
@russell-d-e : We have the following
  " TEST_PROCEDURE is a COLLECTION of TEST_STEP which in turn is a COLLECTION of TEST"
and associated with these we have respectively
  " TEST_LOG, TEST_RECORD and TEST_RESULT"
While the "note" tells us that TEST_LOG is for TEST_PROCEDURE" etc. that may not be something people have in their mental model and they may not have easy access to the "note". 
Q. So, do you think we should rename TEST_LOG and TEST_RECORD?
Q. We have 3 levels in this hierarchy - just checking. You know this much better.

We have the following
	TEST_RECORD (note "a TEST_RECORD is a collection of TEST_RESULTs assoicated with a TEST_STEP") is a type of COLLECTION.
		content of TEST_RECORD only has values of type TEST_RESULT.
		nextRecord (note "The next TEST_RECORD that is the next in a sequence " ) describes TEST_STEP with a single value of type TEST_RECORD.
		logs (note "The next TEST_STEP from which the TEST_RECORD is captured" ) describes TEST_STEP with a single value of type TEST_STEP.
		logs is a type of wasImpactedBy.

- correct the spelling of "assoicated"
- did we want domain of nextRecord to be TEST_RECORD? Do we typically navigate by going to "next test step" and looking at the results (in which case the property name could be "record" instead of "nextRecord") or we go from "result" to "next result"?
- also we have log: TEST_STEP -> TEST_STEP?

kityansiu:
Putting a comment here in case git merge doesn't resolve it. Be sure to remove from TESTING.sadl the following lines:

under TEST
`producedBy (note "TEST_DEVELOPMENT activity(s) this test belongs to") describes TEST with values of type ACTIVITY.`
`producedBy is a type of wasGeneratedBy.`

under TEST_RESULT
`executedBy (note "ACTIVITY(s) that created this test result") describes TEST_RESULT with values of type ACTIVITY.`
`executedBy is a type of wasGeneratedBy.`

russell-d-e:
@kityansiu and @AbhaMoitra 

I think I have merged all the changes back from master into this branch. could you do one more quick check just because the merge got a bit messy. If it looks good we can merge and remove the branch.

RACK-Ontology/OwlModels/.gitignore
```diff
@@ -0,0 +1,40 @@
+/AGENTS.owl
```
tuxji, line 1: Please remove this file.  We don't need it since 1) ignoring OWL files already in git doesn't work, and 2) when the time comes, we'll remove OWL files from git and put *.owl in RACK/.gitignore at the root level.

tuxji (commented):
Looks good to me, although I'll defer to Kit and Abha for approval.  I just wanted to say that I hope we'll stop updating OWL and CDR files soon and remove them from git since the CI jobs automatically generate new OWL and CDR files now (meaning that the files in git don't make it to the rack-box dev image anyway).

tuxji:
@russell-d-e, another suggestion too.  As you said, this branch has a messy git history because of its multiple merges and commits.  When I look at the "Files Changed" tab, I see a pretty small changeset (changes to a handful of SADL files) that could easily fit into a single commit with a single commit message.  GitHub has a dropdown button which allows you to select how to merge the pull request into the main branch: 1) Create a merge commit; 2) Squash and merge; 3) Rebase and merge.  You want to pick 2) Squash and merge, which will squash the branch's messy git history into a single commit.  I've used GitHub's Squash and merge button before and the UI works well.  After you press the button, the multiple commits become one commit and you're shown a text editor pane which allows you to edit the combined commit messages however you want.  You can clean up that commit message so you end up with a single commit with a single commit message that mentions only the changes to the SADL files.

kityansiu (approved):

RACK-Ontology/ontology/HARDWARE.sadl
```diff
@@ -0,0 +1,40 @@
...
+import "http://arcos.rack/PROV-S".
+
+HWCOMPONENT is a type of ENTITY.
+    instantiates (note "What logical component (system) does this physical component instantiate or implement?")
+        describes HWCOMPONENT with values of type ENTITY.
+    componentType (note "Type of component")
+        describes HWCOMPONENT with a single value of type COMPONENT_TYPE.
+	partitions describes HWCOMPONENT with values of type PARTITION.
+
+PARTITION (note "a PARTIION divides a HWCOMPOENTS resources into protected, isolated execution space for software components.") is a type of ENTITY.
```
kquick, line 23: "PARTITION",  "HWCOMPONENTS"

glguy: Fixed in #620

PR #599, opened 2021-11-30
migration update

Ptival:
Updates up to today's state of the ontology.

@glguy I think I fairly represented when a property was removed vs. when it was subsumed by one of `wasGeneratedBy`/`wasAttributedTo`. Could you double-check that the instances I marked as `DeleteProperty` are indeed properties that were removed without a subsumption intent?

### DIFF
--- a/migration/rack/commits/commit40955e24b4e38d45df2ffd0ad8aa47a827a4c72f.py
+++ b/migration/rack/commits/commit40955e24b4e38d45df2ffd0ad8aa47a827a4c72f.py
@@ -1,0 +1,103 @@
+# Copyright (c) 2021, Galois, Inc.
+#
+# All Rights Reserved
+#
+# This material is based upon work supported by the Defense Advanced Research
+# Projects Agency (DARPA) under Contract No. FA8750-20-C-0203.
+#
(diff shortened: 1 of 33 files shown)


migration/rack/commits/commit40955e24b4e38d45df2ffd0ad8aa47a827a4c72f.py
```diff
@@ -0,0 +1,96 @@
...
+            name_space=PROCESS, class_id="PROPERTY", property_id="partiallySupports"
+        ),
+        CreateProperty(name_space=PROCESS, class_id="PROPERTY", property_id="scopeOf"),
+        CreateProperty(
+            name_space=PROCESS, class_id="PROPERTY", property_id="mitigates"
+        ),
+        # PROV-S.sadl
+        CreateProperty(name_space=PROV_S, class_id="ACTIVITY", property_id="goal"),
+        # TESTING.sadl
+        DeleteProperty(name_space=TESTING, property_id="producedBy"),
```
glguy, line 94: Should be a Rename

glguy (approved):
Everything other than the rename change looks great to me

PR #602, opened 2021-11-30
Readme updates corresponding to recent Turnstile changes.

kquick:
Clarified the directory locations for the various examples by
referencing the root directory instead of a relative directory from
the current context (which wasn't always clear).

Removed obsolete CLI readme instructions; replaced with references to
the actual scripts to avoid future loss of synchronization.

Fixed some references to the new turnstile import URL.


kityansiu (approved):
Fantastic! Thank you for this.

PR #605, opened 2021-11-30
Fix a typo in the cli init code.

kquick:
Not sure what other effects this typo might have had, or what else
might need to be tested along with this fix.


Ptival (approved):
Nice catch!

I looked into whether there could be a simple way of generating type stubs for the JSON schemas to avoid this problem, but the way the `jsonschema` library is written, it won't be possible to generate such typings automatically.

glguy (approved):

PR #608, opened 2021-11-30
Use string type for entityURL

glguy:


glguy:
The merge is failing, but it doesn't appear that that is because of this change, and I think we're going to need this change either way.

kityansiu:
@tuxji, does this failing CI mean that PROV-S.owl don't get generated? (Apologies, still wrapping my head around the new workflow.)

tuxji:
@glguy, which CI job failed?  The last two CI jobs on the master branch succeeded / will succeed: <https://github.com/ge-high-assurance/RACK/actions/runs/1526336227> and <https://github.com/ge-high-assurance/RACK/runs/4384879662?check_suite_focus=true>.  The CI job on the vtv_security_ontology branch failed due to a lint check that's already fixed in the master branch: <https://github.com/ge-high-assurance/RACK/runs/4383641923?check_suite_focus=true>.  

@kityansiu, if a CI job fails (on the master branch) when generating the OWL files (or doing anything else, like a lint check), the CI job won't replace the rack-box dev image, so the last successfully built rack-box dev image still will be there.

glguy:
@tuxji I'm just taking about the red `X` next to the commit above.

Screen Shot 2021-12-01 at 9 35 04 AM

tuxji:
Oh, that was a weird fluke failure of GitHub's cache action that I hadn't seen before and hope I won't see again.  For some reason, GitHub failed to cache the files assembled by the cache job so the test and build jobs could use these files.  The test job failed with `tar (child): fuseki.tar.gz: Cannot open: No such file or directory` and the build job failed with `==> docker: chmod: cannot access 'systemctl3.py': No such file or directory`, which are among the files supposed to be cached by the cache job.

PR #609, opened 2021-11-30
Migration updates

Ptival:
Mypy is silly, Pylance is more helpful.


Ptival:
I ended up bundling a bunch of changes together, the high-level picture is:

1. I introduced `SubsumeProperty` as an alternative to `RenameProperty` to capture the intent. At the moment, the operational semantics of migrationg are the same, and it is just a matter of what is displayed to the user or written in the changelog, since describing a subsumption as a renaming must have looked somewhat confusing.
2. I added more recent commits.
3. I removed settings files that are not relevant.
4. I merged the mypy settings into `setup.cfg`.

glguy (approved):
Tracking property subsumption makes a lot of sense, and doesn't imply a sort of "overwrite" of the target.

PR #617, opened 2021-12-03
Remove OWL and CDR files

tuxji:
We now generate new OWL and CDR files with scripts and CI jobs so we
no longer need to store OWL and CDR files in the git repository.
However, we need to bump sadl-eclipse to v3.5.0-20211204 so we can
tell it to change its UID and GID to match the GitHub runner's UID and
GID.  Fix SADL errors with incorrect usage of "must be one of" in
SRI.sadl found by sadl-eclipse bump.

action.yml: Make run commands more uniform.

.gitignore: Ignore OWL and CDR files to avoid adding to git again.

**/OwlModels/*.owl: Remove from git.

GE-Ontology/OwlModels/ont-policy.rdf: Update.

RACK-Ontology/OwlModels/ont-policy.rdf: Update.

SRI.sadl: SADL-Eclipse v3.5.0-20211204 now says that "must be one of"
is incorrectly used on line 239.  The SADL errors go away after
changing line 239 from "must be one of" to "only has values of type".

cli/setup-build-owl.sh: Bump sadl-eclipse to v3.5.0-20211204 and tell
it to change its UID and GID to match the user's own UID and GID.

nodegroups/CDR: Remove from git.

nodegroups/README.md: Explain that the CDR and ingestion
subdirectories are automatically generated.

nodegroups/ingestion: Remove from git.


tuxji:
CI jobs have errors - I'll take a look and figure out what's causing the errors.

GregIthaca:
I might be doing something dumb, but it seems like this change has resulted in the standard scripts like `cli/setup-arcos.sh` no longer working out of the (RACK-in-a) box. If running `cli/setup-owl.sh` first is a prerequisite, I think this needs to be made more clear. 

Also, since we rename our rack instance, the ability to specify the docker image tag but not the docker image name (hardcoded in `cli/setup-owl.sh` to `gehighassurance/rack-box`) is inadequate to make this tool run, and it has to be manually edited to adjust for this. As far as I can tell, it's the only one of the scripts that hard-codes the docker image.

Better yet, you could change the `rack model import` so that it works from the SADL source, if that's how the OWL files are being generated in CI anyway?

PR #626, opened 2021-12-08
setup-owl.sh: Copy generated nodegroups, too

glguy:
This change also removes the trailing /OwlModes/ DEST_DIR from the copy commands as this makes the script behave differently the second time it is run


tuxji (approved):
+1

kityansiu (approved):

kityansiu:
+1

PR #632, opened 2022-01-05
Update Turnstile and ScrapingToolkit for v9 ontology changes

glguy:


russell-d-e (commented):
I reviewed the chances for the Turnstile example scripts and everything looked good to me. I did looked at the Add.py and CONSTANTS.py and I assume they are good. since they are autogenerated.  I also looked through the CSVs and everything looked reasonable.

PR #633, opened 2022-01-10
Add baselines for turnstile requirements

glguy:
@russell-d-e How do you imagine baselines fitting into the turnstile example? My understanding is that we have two versions of the requirements already. Are there other versioned artifacts that we should create baselines for?

Are there other unversioned entities that you could imagine having baselines? I'm thinking even though we only have one version of the software that we could still create a release baseline for it. What else would make sense to have a formal, named release in a project like Turnstile?


russell-d-e:
@glguy . I looked through this pull request you have in this is a really good start for an example of baselines.  Something that I think we should add is what we talked about with baseline hierarchies. So what I think we could add is one SW/Design baselines (v1). The two release baselines for can be created (v1 and v2).  The gist of the this is it would look something like what you might see for a doc only change, so Turnstile Release : v1 would be composed of Requirements Package : v1 and SW Design and Code : v1; Turnstile Release : V1.1 would have Requirements Package : v1.1 and SW Design and Code : v1.   The you could have release 1.1 to be derived from v1. 

I think this would show a pretty good example of how baselines can be combined, refined, and used in hierarchy and with requirements. 

This example could be further expanded by adding baselines associated with Testing. This might be a interesting example in that I believe that the testing is all referencing the v1 requirements; this could be interesting to show the lack of testing evidence associated with v1.1 compared to v1.

Beyond that I think there is a lot of different baselines examples that we could expand to include things like open/closed change drivers, but a lot of these would requirement creating additional data as part of the turnstile example.

glguy:
@russell-d-e I think I've incorporated all of your ideas into the current PR

PR #637, opened 2022-01-14
Add virtualbox support to setup-owl

glguy:
This script will use SSH to extract OWL and CDR files from a running RACK VirtualBox instance given that instance's IP address passed to the `-i` parameter.

The script installs an SSH key-pair onto the VM to facilitate the running of scp commands as I couldn't find any other portable way to avoid repeated prompts for the SSH password.

I prefixed a bunch of outputs with `[setup-owl]` to help differentiate them from outputs that might leak out from other executables along the way.

### DIFF
--- a/cli/setup-owl.sh
+++ b/cli/setup-owl.sh
@@ -55,7 +55,7 @@ in
                         ssh-keygen -q -N "" -f rack_ssh_key
                 fi
 
-                echo "[setup-owl] Please confirm the SSH host key if prompted.
+                echo "[setup-owl] Please confirm the SSH host key if prompted."
                 echo "[setup-owl] Please enter the default password if prompted: $(tput bold)ubuntu$(tput sgr0)"
                 ssh "ubuntu@${virtualbox_ip}" "mkdir -p ~/.ssh/; cat >> ~/.ssh/authorized_keys" < rack_ssh_key.pub
         fi
--- a/cli/setup-owl.sh
+++ b/cli/setup-owl.sh
@@ -57,7 +57,9 @@ in
 
                 echo "[setup-owl] Please confirm the SSH host key if prompted."
                 echo "[setup-owl] Please enter the default password if prompted: $(tput bold)ubuntu$(tput sgr0)"
-                ssh-copy-id -i rack_ssh_key "ubuntu@${virtualbox_ip}"
+
+                # We're using this awkward command instead of ssh-copy-id because some old system don't have ssh-copy-id
+                ssh "ubuntu@${virtualbox_ip}" "mkdir -p ~/.ssh/; cat >> ~/.ssh/authorized_keys" < rack_ssh_key.pub
         fi
 
         echo "[setup-owl] Copying OwlModels"
--- a/cli/setup-owl.sh
+++ b/cli/setup-owl.sh
@@ -11,23 +11,25 @@ rack_tag="v9.0"
 sadl_image="sadl/sadl-eclipse"
 sadl_tag="v3.5.0-20211204"
 
-mode="copy"
+mode="docker"
 
 usage() {
-        echo "Usage: setup-owl.sh [-h] [-b] [-n NAME] [-t TAG]"
+        echo "Usage: setup-owl.sh [-h] [-b] [-n NAME] [-t TAG] [-i IPADDRESS]"
         echo "   -h    print help"
         echo "   -b    build OWL files using sadl-eclipse"
         echo "   -n    specify rack-box docker image name"
         echo "   -t    specify rack-box tag when copying OWL files from running image"
+        echo "   -i    get OWL files from VirtualBox VM by IP address"
         exit 1
 }
 
-while getopts "bn:t:" o; do
+while getopts "bi:n:t:" o; do
         case "$o"
         in
                 b) mode="build";;
                 n) rack_image="${OPTARG}";;
                 t) rack_tag="${OPTARG}";;
+                i) virtualbox_ip="${OPTARG}"; mode="virtualbox";;
                 *) usage;;
         esac
 done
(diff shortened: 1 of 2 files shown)


glguy:
For the record I did run spellcheck on the script before pushing and had it come back without output, but apparently I ran it before making the last set of changes? I'm not sure, but I'm fixing it.

kquick:
> For the record I did run spellcheck on the script before pushing ...

You should run shellcheck as well.  :-)

cli/setup-owl.sh
```diff
@@ -11,34 +11,74 @@ rack_tag="v9.0"
...
+
+                if [ ! -e rack_ssh_key ]
+                then
+                        echo "[setup-owl] Generating a fresh SSH keypair"
+                        ssh-keygen -q -N "" -f rack_ssh_key
+                fi
+
+                echo "[setup-owl] Please confirm the SSH host key if prompted."
+                echo "[setup-owl] Please enter the default password if prompted: $(tput bold)ubuntu$(tput sgr0)"
+                ssh "ubuntu@${virtualbox_ip}" "mkdir -p ~/.ssh/; cat >> ~/.ssh/authorized_keys" < rack_ssh_key.pub
```
tuxji, line 60: I believe it would be easier to use [ssh-copy-id](https://www.ssh.com/academy/ssh/copy-id) instead of this complicated command.

tuxji (approved):
+1

Let's use ssh-copy-id and get this in asap.

cli/setup-owl.sh
glguy: I agree with you, but seem to remember that some systems don't have that command. I've had to install it manually before.

Do you know that it's currently ubiquitous enough to rely upon?

tuxji: Good point.  I don't have a Mac, but this MacAppStore [page](https://macappstore.org/ssh-copy-id/) seems to imply that Macs may not have ssh-copy-id installed by default.  We'd better keep this line as it is instead.

glguy: OK, I've added a comment explain why that ugly command is there for the next reader

cli/setup-owl.sh
```diff
@@ -11,34 +11,74 @@ rack_tag="v9.0"
...
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/Turnstile-Example/Turnstile-IngestionPackage/CounterApplicationUnitTesting/OwlModels" "${rack_dir}/Turnstile-Example/Turnstile-IngestionPackage/CounterApplicationUnitTesting/"
+
+        echo "[setup-owl] Copying nodegroups"
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/nodegroups/CDR" "${rack_dir}/nodegroups/"
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/nodegroups/ingestion" "${rack_dir}/nodegroups/"
+
+        ;;
+
+    docker)
+        echo "[setup-owl] Coping CDR and OWL files from Docker"
```
kquick, line 82: "copying"

cli/setup-owl.sh
```diff
@@ -11,34 +11,74 @@ rack_tag="v9.0"
...
+                        ssh-keygen -q -N "" -f rack_ssh_key
+                fi
+
+                echo "[setup-owl] Please confirm the SSH host key if prompted."
+                echo "[setup-owl] Please enter the default password if prompted: $(tput bold)ubuntu$(tput sgr0)"
+                ssh "ubuntu@${virtualbox_ip}" "mkdir -p ~/.ssh/; cat >> ~/.ssh/authorized_keys" < rack_ssh_key.pub
+        fi
+
+        echo "[setup-owl] Copying OwlModels"
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/RACK-Ontology/OwlModels" "${rack_dir}/RACK-Ontology/"
```
kquick, line 66: Are we intentionally "flattening" the namespace by removing the "OwlModels" subdir?

cli/setup-owl.sh
```diff
@@ -11,34 +11,74 @@ rack_tag="v9.0"
...
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/GE-Ontology/OwlModels" "${rack_dir}/GE-Ontology/"
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/GrammaTech-Ontology/OwlModels" "${rack_dir}/GrammaTech-Ontology/"
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/STR-Ontology/OwlModels" "${rack_dir}/STR-Ontology/"
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/Boeing-Ontology/OwlModels" "${rack_dir}/Boeing-Ontology/"
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/LM-Ontology/OwlModels" "${rack_dir}/LM-Ontology/"
+        scp -q -i rack_ssh_key -r "ubuntu@${virtualbox_ip}:RACK/SRI-Ontology/OwlModels" "${rack_dir}/SRI-Ontology/"
```
kquick, line 72: As a thought (perhaps for the future), it would be great if we could `for dir in $(ssh ubuntu@${virtualbox_ip} find RACK -type d -name OwlModels) ; do scp ...; done`

cli/setup-owl.sh
kquick: Oops, my mistake.  I was seeing there was a trailing `/*` on the source side but that's not actually there.

PR #639, opened 2022-01-20
Command to store a single nodegroup

glguy:


glguy:
@russell-d-e this certainly isn't the end of the nodegroup functionality we discussed, but I wanted to add it first as it sounded like you were hoping to use it.

russell-d-e:
Overall, changes look good, but there is one thing that I am wondering.  Looking through this I think there will be an error if the nodegroup id already exists ( `semtk.store_nodegroup` fails if a nodegroup_id already exists). Should we add some code to remove nodegroup id if it already exists in the store, (similar to how `semtk.store_nodegroups` works).

PR #642, opened 2022-01-22
Add nodegroup store and count to import.yaml

glguy:
With this PR we can add nodegroup storage and checked count operations to the import yaml.

```yaml
ingestion-steps:
  - nodegroup_json: 'nodegroup.json'
    creator: 'Eric'
    comment: 'here we go'
    name: 'file_example'
  - count: 28
    nodegroup: 'file_example'
```

This also lets us use parameterized nodegroups which means we can more easily reuse nodegroups for performing multiple checks

```
ingestion-steps:
  - nodegroup_json: 'nodegroup.json'
    creator: 'Eric'
    comment: 'here we go'
    name: 'section_example'
  - count: 28
    nodegroup: 'section_example'
    constraints:
      - 'identifier=document_id_1'
```
Screen Shot 2022-01-21 at 4 21 12 PM

```
$ rack data import import.yaml
Adding nodegroup section_example                 OK
Counting nodegroup section_example                 FAIL got:0 expected:28
```

ping @russell-d-e

### DIFF
--- a/cli/rack/__init__.py
+++ b/cli/rack/__init__.py
@@ -111,6 +111,16 @@ INGEST_CSV_CONFIG_SCHEMA: Dict[str, Any] = {
                             'comment': {'type': 'string'},
                             'nodegroup_json': {'type': 'string'}
                         }
+                    },
+                    {
+                        'type': 'object',
+                        'additionalProperties': False,
+                        'required': ['count', 'nodegroup'],
+                        'properties': {
+                            'count': {'type': 'number'},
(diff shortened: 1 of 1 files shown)


glguy:
This doesn't add anything you can't already do with shell scripts built using the rack cli, but hopefully as we keep building out this functionality we'll both set the expectations of including these steps in an ingestion package and also make it easy enough that others will actually do it.

russell-d-e (approved):

PR #646, opened 2022-01-28
Scraping Tool Kit fixes

russell-d-e:
Made updates to use CONSTANTS.py to handle situations where the ingestion nodegroup name has invalid characters  for python function calls. ReadNodegroups.py was updated to correctly get the nodegroup name regardless of os.


AbhaMoitra (approved):
The changes look fine to me, and we should test them out in the dev branch.

ScrapingToolKit/README.md
```diff
@@ -7,6 +7,9 @@ STK is installed by downloading the files to your local machine.  This can eithe
 
 Dependencies:
 `pip install lxml`
```
glguy, line 9: for these we could tell people to run:

```
pip install -r requirements.txt
```

russell-d-e: updated readme to tell people to install dependencies via requirements.txt

tuxji:
Is this pull request still relevant?  It should either be merged or closed, not left to linger for a long time.

PR #649, opened 2022-02-01
[convertStandards] .sadl replaced with .csv files

kityansiu:


RACK-Ontology/ontology/DO-178C/SECTION_2.csv
```diff
@@ -0,0 +1,82 @@
...
+,,,A-1,,Software Planning Process,,A-1-5,,,,,,
+,,,A-1,,Software Planning Process,,A-1-6,,,,,,
+,,,A-1,,Software Planning Process,,A-1-7,,,,,,
+,,,A-2,,Software Development Process,,A-2-1,,,,,,
+,,,A-2,,Software Development Process,,A-2-2,,,,,,
+,,,A-2,,Software Development Process,,A-2-3,,,,,,
+,,,A-2,,Software Development Process,,A-2-4,,,,,,
+,,,A-2,,Software Development Process,,A-2-5,,,,,,
+,,,A-2,,Software Development Process,,A-2-6,,,,,,
+,,,A-2,,Software Development Process,,A-2-7,,,,,,
```
Ptival, line 15: Not important for this merge request, but I wonder if we could get away with only ever mentioning "fields" once...

RACK-Ontology/ontology/DO-178C/AGENT_1.csv
```diff
@@ -0,0 +1,3 @@
+description,identifier,title,definedIn_identifier,actedOnBehalfOf_identifier,dataInsertedBy_identifier
+,SC-205,,,,
+,RTCA,,,,
```
Ptival, line 3: Is it fine that both `DO-178C` and `DO-330` define these two agents.

I don't think it's a problem in the store because they go in their respective data graphs, rather I'm asking whether the repetition is okay with us (it's definitely simpler).

RACK-Ontology/ontology/DO-330/OBJECTIVE_2.csv
```diff
@@ -0,0 +1,77 @@
+description,entityURL,generatedAtTime,identifier,invalidatedAtTime,title,definedIn_identifier,satisfiedBy_identifier,dataInsertedBy_identifier,wasAttributedTo_identifier,wasDerivedFrom_identifier,wasGeneratedBy_identifier,wasImpactedBy_identifier,wasRevisionOf_identifier
+"Tool Executable Object Code is produced. ","","","T-2-7","","","","","","","","","",""
```
Ptival, line 2: Not sure why, but this CSV files uses a lot of empty strings for the missing fields, whereas the other ones just put nothing.

This may cause problems?

Ptival (approved):
Alright, it seems like this indeed ports all of the ARP, the DO-178C, and the DO-330 to CSV.

Apart from the remark about the use of empty strings rather than null values, I believe this is good to go!

RACK-Ontology/ontology/DO-330/OBJECTIVE_2.csv
kityansiu: Ah. Good find. Didn't cause any problems, but I cleaned up the empty strings anyway.

PR #653, opened 2022-02-03
Add range restrictions on wasGeneratedBy for well-known generation activities

### LINKED ISSUE #630: How to connect ANALYSIS to ANALYSIS_OUTPUT?
In the ANALYSIS.sadl ontology, an ANALYSIS is a type of ACTIVITY which has a inputs and configuration but no output, so ANALYSIS_OUTPUT isn't connected to this in any obvious way. 

Since an ANALYSIS_OUTPUT inherits from ENTITY, the `wasGeneratedBy` attribute could point to the ANALYSIS. If this is the intent, then should it be specified that `wasGeneratedBy of ANALYSIS_OUTPUT only has values of type ANALYSIS.` in the same way that this is specialized for many other places?

glguy:
Per our discussion this PR identifies ENTITIES that have specific ACTIVITY subtypes that are intended to create them.

Fixes #630

### DIFF
--- a/RACK-Ontology/ontology/ANALYSIS.sadl
+++ b/RACK-Ontology/ontology/ANALYSIS.sadl
@@ -47,3 +47,5 @@ ANALYSIS_OUTPUT
 
     analyzes (note "ENTITY(s) under analysis") describes ANALYSIS_OUTPUT with values of type ENTITY.
     analyzes is a type of wasImpactedBy.
+    
+    wasGeneratedBy of ANALYSIS_OUTPUT only has values of type ANALYSIS.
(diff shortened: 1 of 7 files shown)


STR-Ontology/OwlModels/configuration.rdf
```diff
@@ -2,7 +2,7 @@
     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
     xmlns="http://com.ge.research.sadl.configuration#">
   <Category rdf:about="http://com.ge.research.sadl.configuration#ProjectDependencies">
-    <pDependsOn>file:/RACK/RACK-Ontology</pDependsOn>
+    <pDependsOn>file:/Users/emertens/Source/arcos/RACK/RACK-Ontology</pDependsOn>
```
russell-d-e, line 5: I believe the rdf files are generated by sadl, should/could they be removed from the repo.  I say this because I noticed they have a user specific path (i.e. /Users/emertens...)

russell-d-e (approved):

PR #655, opened 2022-02-03
Remove .rdf generated files from the repository

glguy:
These files contain paths that only make sense on the local computer and are not needed for our tooling.


glguy:
Previously, I believe, these files at absolute paths in them. I can see today they are relative to the project. They will be regenerated when we use SADL to regenerate the OWL files should we ever need them. They are not used by our automation or scripts. I think the same reasons we don't need the OWL files tracked in the repository apply to these files.

tuxji:
Andy Crapo told me that SADL removes old mappings from ont-policy.rdf but doesn't remove the file itself because this file tells SADL where all the imported dependencies come from.  I'm not sure we should remove the .rdf files entirely - from what Andy said below, it sounds like the .rdf files carry important setup information that developers create when using SADL to create a project.

> 1.	It appears that you are completely cleaning the ImplicitModel folder and the OwlModels folder of a project. This is not what SADL does and I expect may have some undesirable effects.
o	In the ImplicitModel folder, the SadlImplicitModel.sadl is supposed to only be replaced if it is missing. This allows a project developer to add to or alter the SadlImplicitModel.sadl file to the needs of the particular project.
o	In the ImplicitModel folder, the SadlBuiltinFuncitons.sadl file is generated from the selected reasoner, identified in the configuration.rdf file. Unless the reasoner is changed there should be no need to regenerate this file each time.
o	In the OwlModels folder, the content of the ont-policy.rdf file gets mappings for OWL files generated from SADL files, from OWL files generated from importing CSV files, and from external OWL models which are added to a project. Unless you are regenerating mappings from all of these sources, you will break the project's ability to find imported OWL models and models generated from CSV files. To my recollection, SADL does not ever delete this ont-policy.rdf. It does attempt to clean old stuff to keep it consistent and relevant, but it doesn't remove it.

glguy:
If there's some value in these generated files then we can close the pr. We don't use them for anything I'm aware of or customize them in any way.

I'm not touching the implicit model sadl files in this pr. We do not customize those generated files either, but I wasn't intending to change them.

glguy:
maybe we'll use these in the future

PR #658, opened 2022-02-08
Auto-add trailing slash for ASSIST -m URL specifications.

kquick:
Automatically supplies a trailing slash if one is missing for detected URL -m specifications.  For example:

```
  check -m http://localhost:3030
```

will automatically get translated to:

```
  check -m http://localhost:3030/
```


glguy (approved):
I ran it locally and it worked with and without the slash. Thanks!

PR #660, opened 2022-02-10
Incorporate MITRE CWE into the core ontology

glguy:
This PR builds on the recent work between our team and GrammaTech. This PR further refines what GT is doing to separate out the identities of detection methods from their effectiveness.

This might be the first place in the ontology that we have a reified relationship class. `CWE_DETECTION_EFFECTIVENESS` exists so that we can have attributes (effectiveness) on the relationship between CWEs and detection methods. I think this is the proper encoding, but we should look at this deliberately.

There are CWEs (e.g. Dead Code) and Detection methods (Manual Static Analysis - Source Code). Detection methods can have different effectivenesses when applied to different CWEs. In this case the effectiveness is listed as High, but in other cases it might be Moderate or "SOAR Partial". The effectivenesses are particular to the pairing.

The example submitted by GrammaTech did not give each detection method a unique identity. Each method existed uniquely instantiated to a CWE. It's possible that we don't need the flexibility, but without giving the detection methods a consistent identity I don't know what we could ever do with them other than print their pairing-specific names out.

```
  detectionImplementedBy describes CWE_DETECTION_METHOD with values of type TOOL.
```

I think we want a way like this to indicate that some tools implement known detection methods. Once we have this we can do queries to find software that has weaknesses that aren't effectively (or are) analyzed. There might be a better way to make this link. Perhaps it should go on the CWE_ANALYSIS because different tools have different operating modes and they aren't always using the same set of detections.


cli/setup-arcos.sh
```diff
@@ -28,3 +28,4 @@ rack nodegroups import ../nodegroups/ingestion/arcos.arbiter
 # rack data import --clear ../RACK-Ontology/OwlModels/ARP-4754A.yaml     # from datagraph http://rack001/arp-4754a
 # rack data import --clear ../RACK-Ontology/OwlModels/DO-330.yaml        # from datagraph http://rack001/do-330
 # rack data import --clear ../RACK-Ontology/OwlModels/DO-178C.yaml         # from datagraph http://rack001/do-178c
+rack data import --clear ../RACK-Ontology/OwlModels/MITRE-CWE.yaml     # from datagraph http://rack001/mitre-cwe
```
GregIthaca, line 31: Eric, this confuses me. Why is it 'rack data import' but pointing to the OwlModels; shouldn't it be 'rack model import'?
Of course there also needs to be a 'rack data import' but that would be using the nodegroups instead, right?

GregIthaca (changes requested):
This all looks reasonable to me other than the one question I rased. As far as I can tell, the CWE data still needs to be added somewhere?

cli/setup-arcos.sh
glguy: This is confusing because it's wrong. I'll fix it

glguy:
I've updated this PR to be complete with a set of data and the ingestion process.

It includes:
- The list of CWEs
- The list of effectiveness enumeration values
- The list of detection methods
- The association between CWEs, methods, effectivenesses, and descriptive text about that association

ping @GregIthaca

glguy:
I've updated this PR to ingest the list of CAPECs and all the wasImpactedBy relations to CWE for them.

Our original CAPEC list is much shorter. It seems better to me to simply list all the known CAPECs than to select from the list, but I'm interested to hear if there's a good case for the more restricted list. It seems like we might just chose not to have some CAPECs not linked to anything from the Boeing data.

kquick (approved):

PR #672, opened 2022-03-09
correcting Turnstile overlay

kityansiu:
- removed turnstile:wasInformedBy and replaced with wasInformedBy (so it doesn't create a new property when it's the one from ACTIVITY that we're after)
- corrected the way the property range was being restricted by adding the term 'only' in SADL


cuddihyge (approved):
These look correct to me.

PR #682, opened 2022-03-18
Update setup-owl CLI utility for rack docker version v10.0.

kquick:


tuxji (approved):
+1

I've also added setup-owl.sh to the list in RELEASE.md so we will know we have to update it next time we make a release.

PR #687, opened 2022-04-06
Add missing CWE entries

glguy:
Previously only the "Software Development" CWEs were loaded. This completes the list to also load "Hardware Design" and "Research Concepts" which covers things like the previously missing CWE-285.

Beyond that this also populates the `title` property in addition to the previously populated `identifier` and `description`


kityansiu (approved):

GregIthaca:
Could I please confirm that the identifiers will have the 'CWE-' prefix from here going forward? We need to match this with our findings, and at one point it was just the number.

kityansiu:
Yes, I can confirm that. We have no plans to change these identifiers.

PR #689, opened 2022-04-06
Add RTX overlay to support synthetic analysis outputs

glguy:


kityansiu (changes requested):
Just 2 recommendations:

- in the sadl file, we've been using the convention uri "http://arcos.projectName/companyName" alias projectName. Can we change line 13 in RTX.sadl to `uri "http://arcos.aace/RTX" alias aace`?
- the directory in nodegroups/ingestion will then become arcos.aace. Can we change line 29 in setup-arcos.sh to `rack nodegroup import ../nodegroups/ingestion/arcos.aace`?

Other than these, I think the changes look good!

kityansiu:
I think we missed updating setup-owl.sh to copy RACK/RTX-Ontology/OwlModels!

glguy:
@kityansiu now fixed on master

PR #696, opened 2022-04-26
fix COTS description

Ptival:
Just noticed this during today's meeting.


tuxji (approved):
Looks good to me and Kit wants this included in v10.2. 

+1

PR #702, opened 2022-05-05
Update __init__.py

russell-d-e:
added CDATA tag to string to objectDataString function. this allows the text to have characters that are normally not valid as part of XML string.  For example "<" is encoded to the xml as "&lt;"


tuxji (approved):
+1

PR #706, opened 2022-05-18
fix rack-migration re. PEP508

Ptival:
It appears the tooling around pyproject.toml has recently decided to check the
project name against PEP508.  This no longer allows freeform text as the
project name, so we change it to be compliant.


kityansiu (approved):
working now! thanks!

kityansiu (approved):
working! i'm now about to run rack_migrate.

Ptival:
Good to know. I also have not tested the migration tool in a long time!

russell-d-e (approved):

PR #728, opened 2022-06-22
Update turnstile data generation to use new automatic CDRs

glguy:


cuddihyge (approved):
Thanks Eric.   This is not only the needed upgrade, but a nice example to help us see how to upgrade a couple other pieces of the system.   @kityansiu

PR #746, opened 2022-07-25
Add an import manifest format/command

glguy:
This adds a command: `rack manifest import manifest.yaml` that can do data, model, and nodegroup ingestion from a YAML file like:

```yaml
steps:
  - model: models/import.yaml
  - data: dat1/import.yaml
  - data: dat2/import.yaml
  - nodegroups: a/nodegroup/dir
```


glguy:
I'm planning to add things like baseUrl rewriting into the data command and data import.yaml format instead of putting them into the manifest.

Ping @weisenje @cuddihyge

PR #749, opened 2022-08-04
Stk autogeneration

russell-d-e:
merging STK changes to use the class based ingestions.


kityansiu (approved):
I've tested the changes to the Scraping Toolkit and the Turnstile Ingestion.

PR #752, opened 2022-08-08
Add RACK UI

weisenje:
This is the initial rev of the RACK UI.


.github/workflows/actions/download/action.yml
```diff
@@ -64,6 +64,13 @@ runs:
         cd ${{ github.workspace }}
         tar cfz RACK/rack-box/files/rack-cli.tar.gz RACK/cli/{*.sh,wheels}
 
+    - name: Package RACK UI
+      shell: bash
+      run: |
+        cd RACK/rack-ui
+        cd ${{ github.workspace }}
```
tuxji, line 71: The first cd command on line 70 is redundant and can be removed because the second cd command on line 71 changes the directory again.

.github/workflows/continuous.yml
```diff
@@ -201,18 +201,35 @@ jobs:
...
+      if: github.ref == 'refs/heads/dev-test'
+      run: |
+        cd RACK/rack-box
+        packer build -var version=dev-test rack-box-docker.json
+
+    - name: Build rack-box other image
+      if: github.ref != 'refs/heads/master' && github.ref != 'refs/heads/dev-test'
+      run: |
+        cd RACK/rack-box
+        packer build -var version=other rack-box-docker.json
```
tuxji, line 220: My motivation for writing this step is to allow any feature or fix branch with a different name than "master" or "dev-test" to test that it can build a rack-box Docker image successfully, even though we push only the "dev" and "dev-test" Docker images to Docker Hub in later steps.

I'm not sure if you want to retain the dev-test branch for future work.  Anytime you or someone else want to continue working on the dev-test branch after some commits have been pushed to the master branch, you will have to remember that you must run `git rebase origin/master` to bring the dev-test branch up to date with the master branch before you append any new commits to the dev-test branch.  You may want to remove the dev-test branch and simply recreate the dev-test branch from the master branch whenever you need to use it.  Note if you decide you will never use the dev-test branch again, then you don't need to change this workflow (all of the extra lines or extra steps or changed lines were made to support the dev-test branch).

rack-box/scripts/install.sh
```diff
@@ -70,6 +72,16 @@ cp /opt/fuseki/fuseki.service /etc/systemd/system/fuseki.service
...
+# Set up and start RACK UI service
+
+cd /home/"${USER}"/RACK/rack-ui
+python3 -m pip install -r ./requirements.txt
+adduser --system --group --no-create-home --disabled-password rackui
+mkdir /etc/rackui
+chown rackui.rackui /etc/rackui
+envsubst < rackui.service > /etc/systemd/system/rackui.service
+systemctl enable rackui
+
```
tuxji, line 84: How are you starting the rack ui if it needs to be started?  Do you need to add `systemctl start rackui` at line 84 or do you want users to start it manually?

rack-ui/app.py
```diff
@@ -0,0 +1,209 @@
...
+#             with redirect_stdout(in_mem_file):
+#                 # load RACK
+#                 rack.ingest_manifest_driver(Path("../cli/rack.yaml"), BASE_URL, TRIPLE_STORE, TRIPLE_STORE_TYPE, True)
+#             return True, "RACK has been reset"  # show modal dialog
+#         except Exception as e:
+#             return True, get_error_trace(e)  # show modal dialog with error
+#     elif modal_is_open and n_clicks_close is not None and n_clicks_close > 0:  # user clicked close on modal dialog
+#         in_mem_file.truncate(0)  # clear status
+#         in_mem_file.seek(0)  # clear status
+#     return False, ""  # hide (or don't show) modal dialog
```
tuxji, line 181: If you don't plan to uncomment this code later, best practice is remove all dead code to avoid cluttering greps hits, searches, etc.

rack-ui/rackui.service
```diff
@@ -0,0 +1,15 @@
...
+Type=simple
+WorkingDirectory=/home/${USER}/RACK/rack-ui
+ExecStart=/usr/bin/python3 app.py
+TimeoutStopSec=20
+KillMode=process
+User=rackui
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
```
tuxji, line 15: Should add a final newline.

tuxji (approved):
+1 (after you address comments below)

GitHub reports a conflict between this PR and the master branch.  When you rebase your dev-test branch on the master branch to resolve that conflict, I think it would be a good idea to squash your commits down to a single commit as well.  Otherwise the rebase will report a conflict somewhere in one of the 65 commits, pause, wait for you to fix the conflict before resuming rebasing, probably report another conflict in another commit, pause again, etc.

You have to use `git fetch && git rebase -i origin/master` and manually edit the list of 65 commits in a text editor to squash your commits down to 1 commit.  The first commit should remain "pick" and the rest of the commits should be changed from "pick" to "squash" or just "s".  Git will append all the commits' messages into a single commit message and let you edit that commit message to clean it up before completing the squash.  

The difference between the above command and `git fetch && git rebase origin/master` (without the "-i") is that the latter command will rebase all 65 commits one after another, and you probably will have to fix conflicts and continue rebasing several times.  It's simpler to squash the commits together into a single commit so that you'll have to fix any conflicts in that commit only once.

.github/workflows/continuous.yml
tuxji: That's right, the if statements should not break if the dev-test branch is missing.

PR #778, opened 2022-09-21
Compile SADL files

tuxji:
Running "cli/setup-owl.sh -b" on RACK's SADL files produces the following changes in addition to generating OWL files (which aren't checked into git):

modified: GE-Ontology/OwlModels/ont-policy.rdf
new file: sadl-examples/OwlModels/TurnstileSecurity.rules
new file: sadl-examples/OwlModels/configuration.rdf
new file: sadl-examples/OwlModels/ont-policy.rdf

We might as well check these files into git so that anyone else running setup-owl.sh in build mode doesn't end up with modified or untracked files.  Agreed?


glguy:
Should we commit them or add them to .gitignore? We're already ignoring the other `*.owl` files as they are output artifacts from this process. I'm guessing you already considered .gitignore and these files perhaps contain some valuable semantic content and aren't just outputs of compilation? In that case I'm in agreement that we should merge this.

Incidentally I think that with the new approach of putting OWL files into the ingestion packages directly that none of our users will have to run this script any more. We still benefit from being able to run the script in CI so I wouldn't want to get rid of it.

kityansiu:
i don't see any harm in checking in these files. i'm okay with that decision.

kityansiu (approved):

PR #781, opened 2022-09-29
Fix ASSIST-DV to handle owl:allValuesFrom restriction.

kquick:
Previously handled owl:someValuesFrom, but it did not identify a property_extra value_from constraint for owl:allValuesFrom.

Also added some convenience functions to help/simplify the higher level check.

Fixes issue #780


Ptival (approved):
I don't think I quite get all the changes to `check_target_type`, rest seems reasonable.

glguy (approved):

PR #788, opened 2022-10-05
[cli] Update README "how-to-use" section with new manifest command.

kquick:


glguy (approved):
My only comment would be that manifest import subsumes all the other data, model, and nodegroup imports, rather than being specific to data.

PR #794, opened 2022-10-12
change to ingestion by class on NIST, CWE, and ARCOS data

kityansiu:


manifests/arcos.yaml
```diff
@@ -11,24 +11,19 @@ steps:
...
 
   - model: ../STR-Ontology/OwlModels/import.yaml
-  - nodegroups: ../nodegroups/ingestion/arcos.arbiter
 
   - model: ../RTX-Ontology/OwlModels/import.yaml
 
   - data: ../RACK-Ontology/ontology/NIST-800-53/import.yaml
 
-  - model: ../RACK-Ontology/OwlModels/MITRE-CWE.yaml
+  - model: ../RACK-Ontology//OwlModels/MITRE-CWE.yaml
```
Ptival, line 27: This looks unnecessary.

Ptival (commented):
All classes look correct.

What is the intent behind the removal of the nodegroups in `arcos.yaml`?

manifests/arcos.yaml
kityansiu: yikes. typo.

kityansiu: I just wanted to make it lighter weight, so we don't have to load all these nodegroups into RACK-in-a-Box if we are not relying on them for ingestion.

PR #806, opened 2022-10-20
adding vim to Docker

Ptival:
I hear rumors that some other TAs would appreciate vim being in our Docker images out of the box, and I concur.

Unless we have a reason not to, I propose we just add it.

@tuxji Any reason not to do this?


tuxji (approved):
+1

For consistency, please also add vim to the list of packages in RACK/rack-box/http/user-data.

PR #811, opened 2022-10-27
add optimize.sh and set up Docker with tdbstats

Ptival:
First draft for an "optimize" script that runs tdbstats.

We want to iterate on this to make it work in more settings, but this is a start for Docker.

### DIFF
--- a/.github/workflows/actions/download/action.yml
+++ b/.github/workflows/actions/download/action.yml
@@ -29,6 +29,11 @@ runs:
       run: |
         curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-fuseki-4.6.1.tar.gz -o RACK/rack-box/files/fuseki.tar.gz
 
+    - name: Download Apache Jena libraries
+      shell: bash
+      run: |
+        curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-4.6.1.tar.gz -o RACK/rack-box/files/jena.tar.gz
+
     - name: Download SemTK release
       shell: bash
       run: |
--- a/rack-box/README.md
+++ b/rack-box/README.md
@@ -39,6 +39,9 @@ although we will mention each file here as well:
 - `files/fuseki.tar.gz`: Download latest Fuseki release tarball from
   <https://jena.apache.org/download/> and rename it
 
+- `files/jena.tar.gz`: On the same page you downloaded Fuseki, scroll down and
+  get the Apache Jena libraries, renaming it to `jena.tar.gz`
+
 - `files/semtk.tar.gz`: Download latest SemTK release tarball from
   <https://github.com/ge-semtk/semtk/releases> and rename it
 
--- a/rack-box/scripts/install.sh
+++ b/rack-box/scripts/install.sh
@@ -54,6 +54,9 @@ mkdir -p "/home/${USER}"
 tar xfzC fuseki.tar.gz /opt
 rm fuseki.tar.gz
 mv /opt/apache-jena-fuseki-* /opt/fuseki
+tar xfzC jena.tar.gz /opt
+rm jena.tar.gz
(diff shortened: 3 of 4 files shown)


.github/workflows/actions/download/action.yml
```diff
@@ -29,6 +29,11 @@ runs:
       run: |
         curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-fuseki-4.5.0.tar.gz -o RACK/rack-box/files/fuseki.tar.gz
 
+    - name: Download Apache Jena libraries
+      shell: bash
+      run: |
+        curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-4.5.0.tar.gz -o RACK/rack-box/files/apache-jena.tar.gz
```
tuxji, line 35: We currently use Fuseki 4.6.1, not 4.5.0, so you should use 4.6.1 as well.  You also might as well shorten the name to jena.tar.gz since we drop the apache- from fuseki.tar.gz too.

rack-box/README.md
```diff
@@ -40,6 +40,10 @@ although we will mention each file here as well:
   <https://jena.apache.org/download/> and rename it (note we still are
   using version 3.16.0 instead of the latest release, though)
 
+- `files/apache-jena.tar.gz`: On the same page you downloaded Fuseki, scroll
```
tuxji, line 43: We've changed lines 39-41 in the master branch.  You should update lines 43-45 to match as well:

```
- `files/jena.tar.gz`: Download latest Jena release tarball from
  <https://jena.apache.org/download/> and rename it
```

rack-box/scripts/install.sh
```diff
@@ -51,6 +51,9 @@ mkdir -p "/home/${USER}"
 tar xfzC fuseki.tar.gz /opt
 rm fuseki.tar.gz
 mv /opt/apache-jena-fuseki-* /opt/fuseki
+tar xfzC apache-jena.tar.gz /opt
+rm apache-jena.tar.gz
```
tuxji, line 55: Shorten name to jena.tar.gz

tuxji (approved):
+1 with 2 suggestions

.github/workflows/actions/download/action.yml
Ptival: Oh it is already changed... Weird, I didn't get these changes. Will fix!

.github/workflows/actions/download/action.yml
```diff
@@ -27,7 +29,12 @@ runs:
     - name: Download Fuseki release
       shell: bash
       run: |
-        curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-fuseki-4.6.1.tar.gz -o RACK/rack-box/files/fuseki.tar.gz
+        curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-fuseki-${FUSEKI_VERSION}.tar.gz -o RACK/rack-box/files/fuseki.tar.gz
+
+    - name: Download Apache Jena libraries
+      shell: bash
+      run: |
+        curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-${FUSEKI_VERSION}.tar.gz -o RACK/rack-box/files/jena.tar.gz
```
tuxji, line 37: You need to tell GitHub CI explicitly when you use an environment variable because run steps can be executed on many shells (even PowerShell).  So you don't write `${FUSEKI_VERSION}`, you write `${{ env.FUSEKI_VERSION }}` instead.  I doubt the env var is worth the effort - it may be easier to just put the 4.6.1 back in (the two lines are close together visually, and nothing should break if the versions accidentally diverge anyway).

rack-box/README.md
tuxji: You still need to update these lines.

PR #827, opened 2022-11-18
Update docker run instructions for new RACK UI port specification.

kquick:


rack-box/Docker-Hub-README.md
```diff
@@ -17,7 +17,7 @@ If you do see these resource settings, make the following changes:
 Now you are ready to start your RACK box.  Type the following command to run your RACK box on your computer:
 
 ```shell
-docker run --detach -p 3030:3030 -p 8050:8050 -p 8080:80 -p 12050-12091:12050-12091 gehighassurance/rack-box:v11
+docker run --detach -p 3030:3030 -p 8050:8050 -p8050:8050 -p 8080:80 -p 12050-12091:12050-12091 gehighassurance/rack-box:v11
```
tuxji, line 20: The -p8050:8050 is redundant because the previous option is -p 8050:8050.

tuxji (commented):
-1

rack-box/Docker-Hub-README.md
kquick: Ha!  Looks like we both fixed it and I didn't pay enough attention when merging my commit with yours.  I reverted this change, but this command line is getting unwieldy anyhow, so I've updated this PR to add `./cli/docker_start.sh`.  I left the original instructions as well since I don't know about the Windows situation, but the script should help for Unix or MacOS users, and be easier for us to avoid additional complexity for the user.

tuxji (commented):
+0

Yes, a script which hides all the port options makes sense.  You'll have to fix the shellcheck warnings first, though.

cli/docker_start.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2086,SC2128,SC2048
```
tuxji, line 2: Was this really the best way to fix the shellcheck warnings?  The changes it recommends usually make your script safer, but I will trust that you made the right decision.

tuxji (approved):
+1

cli/docker_start.sh
kquick: The shellcheck concerns are highly conservative to cover all possible scenarios; in this case much of the script is dealing with output from docker, which is already well-formed and so many of these concerns didn't apply.  One warning was a case where it simply wanted a newline added to fit its preferred layout (not an error, simply a preference of the check author).  The final warning was concerned about a specific usage of an array variable which is a reasonable warning in general, but my use was specific and intentional and thus the warning was not necessary.

PR #834, opened 2022-11-21
Modify core ontology to require an identifier for all THINGs.

kquick:
It seems that we should require an `identifier` for every `THING`, rather than the current state where it is optional.  All objects are identified by a SemTK-generated UUID-based URL, which is not very useful without an associated user-presentable `identifier`.


glguy:
I'm generally in favor of this, but I can think of ways to ingest things that don't use identifiers using custom nodegroups. It's probably the case that the whole current ingestion package has an identifier for everything. Setting this would help us keep it that way.

kityansiu:
This sounds okay to me as well. I added @cuddihyge as a reviewer just to be sure.

kquick:
Ping @cuddihyge

kityansiu (approved):

glguy (approved):
My primary concern here was that the CLAIMS ontology has ENVIRONMENT_RANGE and CONERN, which might not have an obvious identifier, but realistically they should have *an* identifier, whatever we come up with, so that users have a reasonable chance to interact with these things, point to them, associate with them, delete them. Without this interacting with them would require doing a query to list the ones that are available transiently using their randomly assigned URIs, which probably isn't reliable.

kityansiu:
I concur -- that was exactly the problem I accidentally created when I did TEST_RESULTs, which didn't have obvious identifiers, but then made it unreasonably hard to query out to TEST_STATUS.

PR #841, opened 2022-11-24
don't use tables for layout

Ptival:
This gives more or less the same layout, but without resorting to table-based layout.


cuddihyge:
@Ptival and @weisenje can merge or close soon after v12

Issue #823, opened 2022-11-14
RACK-UI: add /optimize endpoint


cuddihyge:
Add an endpoint to the RACK-UI that will run the Fuseki optimization script.
- [x] demonstrate that it's possible to add an endpoint to a Dash app **(Jenny)**
- [x] determine if runs long enough to need asynchronous call (synchronous is fine, biggest dataset finishes in <5 sec) **(Jenny)**
- [x] coordinate with Eric what the endpoint should return **(Jenny)**
- [x] make the endpoint run cli/optimize.sh **(Val)
- [x] test in Docker and VM (not supported on Windows local deployment)

glguy:
That seems fine to me; I don't know what other outputs we'd know to report.

Ptival:
As @glguy mentioned by email, I believe some possible options are:

- Giving more rights to `rackui`, which sounds a bit like giving up entirely on separation,

- Making `root` own `optimize.sh` and giving it a `setuid`. In a private virtual machine, this is pretty innocuous. In a shared setting, this could enable denial-of-service if people are able to run the script repeatedly.

- Otherwise, we would need to come up with some intermediate solution, where the UI is able to request that the script is run, and some process with the appropriate permissions is able to perform that request on its behalf.

@tuxji Do you have any ideas in this space? Are we already doing something similar?

tuxji:
I have the following thoughts:
- When you replace a fuseki file, you want the file to remain owned by the fuseki user
- Therefore, you want to run cli/optimize.sh as fuseki, not as root
- The rack-ui app currently runs as the rackui user but could run as fuseki instead
- Systemd allows you to run executables or shell scripts as any user you wish
- However, you still need to be root to run explicit systemctl start/stop commands
- Systemd can run a script after a service is restarted or trigger systemd services based on conditions

I think your best approach would be to find a way to make systemd run your script when you want it to, or even easier (if it works), run all your commands as the fuseki user including stopping and starting the fuseki server with the fuseki distribution's own [script](https://github.com/apache/jena/blob/main/jena-fuseki2/apache-jena-fuseki/fuseki).  That is, try the following first:

- edit cli/optimize.sh
- replace "systemctl stop fuseki" with "/opt/fuseki/fuseki stop"
- replace "systemctl start fuseki" with "/opt/fuseki/fuseki start"
- edit rack-ui/rackui.service
- replace User=rackui with User=fuseki

[...]

PR #849, opened 2022-12-08
candidate solution for optimize endpoint (#823)

Ptival:
Still working on checking the Virtualbox version.

@tuxji I don't quite know how `http/user-data` comes into play, does it make sense to copy the line that allows user `rackui` to `sudo` without password there?

Should I remove the line from `install.sh` if it's there?

### DIFF
--- a/rack-box/http/user-data
+++ b/rack-box/http/user-data
@@ -27,8 +27,10 @@ autoinstall:
     - python3
     - python3-pip
     - strace
+    - sudo
     - swi-prolog
     - unzip
     - vim
   late-commands:
     - echo 'ubuntu ALL=(ALL) NOPASSWD:ALL' > /target/etc/sudoers.d/ubuntu
+    - echo 'rackui ALL=(ALL) NOPASSWD:ALL' >> /target/etc/sudoers.d/ubuntu
--- a/rack-box/scripts/install.sh
+++ b/rack-box/scripts/install.sh
@@ -28,6 +28,7 @@ apt-get install -yqq \
         python3 \
         python3-pip \
         strace \
+        sudo \
         swi-prolog \
         unzip \
         vim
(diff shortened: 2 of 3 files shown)


rack-box/http/user-data
```diff
@@ -27,8 +27,10 @@ autoinstall:
     - python3
     - python3-pip
     - strace
+    - sudo
     - swi-prolog
     - unzip
     - vim
   late-commands:
     - echo 'ubuntu ALL=(ALL) NOPASSWD:ALL' > /target/etc/sudoers.d/ubuntu
+    - echo 'rackui ALL=(ALL) NOPASSWD:ALL' >> /target/etc/sudoers.d/ubuntu
```
tuxji, line 36: To answer your question whether we really need to modify user-data, the answer is that we used to build a virtual machine from an ISO for which we needed this user-data file.  However, the latest Ubuntu 20.04.5 point release ISO no longer boots and runs the install successfully, so I had to build the virtual machine from an existing virtual machine (a recent Ubuntu vagrant box) instead.  It still would be good to keep user-data around and up to date in case we need to build a virtual machine from an ISO again, so this change is good (although sudo is already installed by the ISO and doesn't need to be listed here).

I suggest changing the last line to create a different and new file to reduce conflict with install.sh:

`    - echo 'rackui ALL=(ALL) NOPASSWD:ALL' > /target/etc/sudoers.d/rackui`

rack-box/scripts/install.sh
```diff
@@ -90,6 +91,8 @@ systemctl start fuseki
 cd /home/"${USER}"/RACK/rack-ui
 python3 -m pip install -r ./requirements.txt
 adduser --system --group --no-create-home --disabled-password rackui
+usermod -aG sudo rackui
+echo "rackui ALL=(ALL) NOPASSWD:ALL" | tee -a /etc/sudoers
```
tuxji, line 95: Likewise, I suggest changing line 95 to create a different and new file to reduce conflict with user-data:

`echo 'rackui ALL=(ALL) NOPASSWD:ALL' > /etc/sudoers.d/rackui`

That way, both user-data and install.sh will end up creating the same file with the same contents anyway.

tuxji (approved):
+1 with a couple suggestions

PR #853, opened 2022-12-08
Update remaining data import.yaml to use class uris

glguy:


glguy:
@weisenje this should address the ingestion issue you were seeing (once it gets merged into the ingestion package you were trying). I think these were the last holdouts for the old ingestion nodegroups.

PR #854, opened 2022-12-09
Add manifest settings for postimport actions

glguy:
Manifests can now specify any of three optional boolean flags:

copy-to-default-graph: true
perform-entity-resolution: true
perform-triplestore-optimization: true

If enabled, these operations run in the order given above.


glguy:
@weisenje this PR includes the manifest functions I believe you're blocked on

Ptival (approved):

PR #858, opened 2022-12-13
Add call to rackui's optimization endpoint

glguy:
Add support for #823 to cli

### DIFF
--- a/cli/rack/__init__.py
+++ b/cli/rack/__init__.py
@@ -427,7 +427,7 @@ def ingest_manifest_driver(
         if manifest.getPerformOptimization():
             invoke_optimization(optimization_url)
 
-def invoke_optimization(url: Optional[Url]):
+def invoke_optimization(url: Optional[Url]) -> None:
     @with_status(f'Optimizing triplestore')
     def go() -> None:
         url = url or DEFAULT_OPTIMIZE_URL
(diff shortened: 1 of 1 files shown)


cli/rack/__init__.py
```diff
@@ -423,7 +425,17 @@ def go() -> dict:
             go()
         
         if manifest.getPerformOptimization():
-            logger.warning("Optimization requested but not yet implemented")
+            invoke_optimization(optimization_url)
+
+def invoke_optimization(url: Optional[Url]) -> None:
+    @with_status(f'Optimizing triplestore')
+    def go() -> None:
+        url = url or DEFAULT_OPTIMIZE_URL
```
Ptival, line 433: Is this fundamentally different than putting a default value on line 430?

cli/rack/__init__.py
```diff
@@ -423,7 +425,17 @@ def go() -> dict:
...
+        response = requests.get(url).json()
+        if not response['success']:
```
Ptival, line 435: Just to double-check, the `'success'` field is populated (with something falsy) even on failure?

Ptival (approved):
Just the 2 nitpicks.

cli/rack/__init__.py
glguy: Yes, it appears so

cli/rack/__init__.py
glguy: I did it because it would mean that the rack ui wouldn't break without some fixup since then I could default the argument to None... I don't know if that was the right call, but it's why I did it.

PR #875, opened 2023-01-13
Add RACK UI utility tab with button to restart Fuseki

weisenje:
@glguy, @Ptival, @cuddihyge - to enable this I added `cli/restart-fuseki.sh` which gets called from the RACK UI, following the same pattern as `cli/optimize.sh`

### DIFF
--- a/EntityResolution/CheckBar.py
+++ b/EntityResolution/CheckBar.py
@@ -5,15 +5,16 @@ from tkinter import *
 class Checkbar(Frame):
     def __init__(self, parent=None, picks=[], side=LEFT, anchor=W,  command=None):
         Frame.__init__(self, parent)
+        self.command = command
         self.vars = {}
+        self.buttons = []
         for pick in picks:
-            var = IntVar()
-            chk = Checkbutton(self, text=pick, variable=var,  command=command)
-            chk.pack(side=side, anchor=anchor, expand=YES)
-            self.vars[pick]=var
-            self.vars[pick].set(1)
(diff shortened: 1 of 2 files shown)


cli/restart-fuseki.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+# Copyright (c) 2022, General Electric Company and Galois, Inc.
+
+set -eu
+
+echo "Stopping Fuseki"
+systemctl stop fuseki
+
+echo "Restarting Fuseki"
+systemctl start fuseki
```
glguy, line 10: Is this doing something that `systemctl restart` doesn't do for us?

rack-ui/pages/utility.py
```diff
@@ -0,0 +1,71 @@
...
+        (Output("restart-button", "disabled"), True, False),        # disable the button while running
+    ],
+    prevent_initial_call=True
+)
+def run_restart(n_clicks):
+    """
+    Restart the triple store
+    """
+    try:
+        if platform.system() == "Windows":
```
glguy, line 47: It seems like we should probably only attempt to restart fuseki when RACK UI is running in our docker image. If the OS is Windows, then obviously we aren't running in our docker image, but a user running on MacOS or Linux running this command probably shouldn't be able to restart (or try to restart) fuseki this way. It likely won't work because they aren't using systemctl to run it (I'm not, for example) or because sudo will prompt for a password and will stall.

Is there some way we can distinguish if we're on Docker to avoid surprises here?

glguy (approved):

PR #877, opened 2023-01-18
Initial support for manifest build

glguy:


cli/rack/__init__.py
```diff
@@ -350,12 +351,162 @@ def utility_copygraph_driver(base_url: Url, triple_store: Optional[Url], triple_
...
+        to_path: Path,
+    ) -> None:
+
+        with open(from_path, mode='r', encoding='utf-8-sig') as f:
+            obj = yaml.safe_load(f)
+
+        base_path = from_path.parent
+        for i in range(len(obj['steps'])):
+            step = obj['steps'][i]
+            if 'manifest' in step:
```
kquick, line 458: FYI, you can replace the above two lines with:

`for (i, step) in enumerate(obj['steps']):`

cli/rack/__init__.py
```diff
@@ -350,12 +351,162 @@ def utility_copygraph_driver(base_url: Url, triple_store: Optional[Url], triple_
...
+                path = step['manifest']
+                dirname = Path(f'{self.next_fresh():02}_manifest')
+                subdir = to_path.parent.joinpath(dirname)
+                topath = subdir.joinpath(Path(path).name)
+                mkdir(subdir)
```
kquick, line 463: Here I think you could use `subdir.mkdir(parents=True, exist_ok=False)` which would also help validate that the `next_fresh`-drived `dirname` was unique.

cli/rack/__init__.py
```diff
@@ -350,12 +351,162 @@ def utility_copygraph_driver(base_url: Url, triple_store: Optional[Url], triple_
...
+                self.data(base_path.joinpath(path), topath)
+                step['data'] = str(dirname.joinpath(Path(path).name))
+            elif 'nodegroups' in step:
+                path = step['nodegroups']
+                dirname = Path(f'{self.next_fresh():02}_nodegroups')
+                subdir = to_path.parent.joinpath(dirname)
+                mkdir(subdir)
+                self.nodegroups(base_path.joinpath(path), subdir)
+                step['nodegroups'] = str(dirname)
```
kquick, line 488: Lots of duplicated if clauses above.  One possible alternative:

```
for attr in [ 'manifest', 'model', ''data', 'nodegroups' ]:
    if attr in step:
        path = step[attr]
        dirname = Path(f'{self.next_fresh():02}_{attr}')
        subdir = to_path.parent.joinpath(dirname)
        subdir.mkdir(parents=True, exist_ok=False)
        topath = subdir.joinpath(Path(path).name)
        # nodegroups is different for the following two; if that's intended, use if attr == "nodegroups" just here...
        getattr(self, attr)(base_path.joinpath(path), topath))
        step[attr] = str(dirname.joinpath(path).name)
        break   # [optional]
```

cli/rack/__init__.py
glguy: These things are being created in a fresh temporary directory. If the parent wasn't created there were be a bug in the program, so I'll leave that one out.

cli/rack/__init__.py
glguy: I agree that there's a bit of duplication here. I'm hesitant to try and factor out more of the commonality because I think the future holds more diversity in ingestion steps (if there's new ones to come) than less, and I don't want to make the code more complicated just to try and eliminate some of the duplication.

cli/rack/__init__.py
```diff
@@ -350,12 +350,169 @@ def utility_copygraph_driver(base_url: Url, triple_store: Optional[Url], triple_
...
+        files = obj['files']
+        for (i,file) in enumerate(files):
+            file = Path(file)
+            shutil.copyfile(frombase.joinpath(file), tobase.joinpath(file.name), follow_symlinks=True)
+            files[i] = file.name
+
+        c = obj.get('model-graphs')
+        if c is None:
+            self.model_graphs.add(str(MODEL_GRAPH))
+        elif type(c) == str:
```
kquick, line 391: Python code usually uses `isinstance` rather than a direct type comparison in order to accomodate subclasses.  For example, if yaml decided to subclass str at some point to add metadata about source location or something like that.

kquick (approved):
A small python idiom comment, but otherwise looks good to me.

PR #890, opened 2023-02-09
Add control flow to the ontology

glguy:
```
/// SAMPLE INSTANTIATION

/*
fun1() { if (fun2()) fun3() }
fun2() { fun3() }
fun3() {}
 */

Fun1 is a SWCOMPONENT.
Fun2 is a SWCOMPONENT.
Fun3 is a SWCOMPONENT.

Acert_v5 is a TOOL.
AcertRun44 is an ACTIVITY wasAssociatedWith Acert_v5.

Fun-1-2-ControlFlow is an AcertControlFlow
  identifier "fun1-fun2-control-flow"
  unconditionalFlow true
  sw:source Fun1
  sw:target Fun2
  wasGeneratedBy AcertRun44. 

Fun-1-3-ControlFlow is an AcertControlFlow
  identifier "fun1-fun3-control-flow"
  unconditionalFlow false
  sw:source Fun1
  sw:target Fun3
  wasGeneratedBy AcertRun44.

Fun-2-3-ControlFlow is an AcertControlFlow
  identifier "fun2-fun3-control-flow"
  unconditionalFlow true
  sw:source Fun2
  sw:target Fun3
  wasGeneratedBy AcertRun44.
```

### DIFF
--- a/GrammaTech-Ontology/ontology/GrammaTech.sadl
+++ b/GrammaTech-Ontology/ontology/GrammaTech.sadl
@@ -49,3 +49,11 @@ Timeout         is a FailureReason with identifier "Timeout".
 Crash           is a FailureReason with identifier "Crash".
 DifferentOutput is a FailureReason with identifier "DifferentOutput".
 Assert          is a FailureReason with identifier "Assert".
+
+AcertControlFlow
+  (note "Control flows computed by the ACERT tooling annotated with condition analysis")
+  is a type of CONTROL_FLOW.
+
+unconditionalFlow
+  (note "The ACERT tooling determined that this control flow is guaranteed")
+  describes AcertControlFlow with a single value of type boolean.
--- a/RACK-Ontology/ontology/SOFTWARE.sadl
+++ b/RACK-Ontology/ontology/SOFTWARE.sadl
@@ -121,3 +121,9 @@ SWCOMPONENT
 
     instantiates (note "What logical component (system) does this physical component (code) instantiate or implement?")
         describes SWCOMPONENT with values of type ENTITY.
+
+CONTROL_FLOW
+    (note "Control flows between software components. Specific flow conditions are captured in subtypes")
+    is a type of ENTITY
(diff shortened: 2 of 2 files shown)


GregIthaca:
I've been drawing up a Boeing example but I have been assuming we would be subclassing SWCOMPONENT and simply restoring two new attributes. I was going to run this by Doug and ask him to direct it to people who have the right agreements in place to review it.

There may be a reason you prefer the approach outlined above, but it feels heavy to me. Here are the tradeoffs I'm seeing:
- This adds 1 entity and 5 triples (?) for each flow, rather than 0 entities and 1 triple for each flow.
- The wasGeneratedBy information is effectively redundant, because the same tool/execution is adding the binary software components and their flows.
- The distinction between conditional and unconditional means adding a "textual" constraint to searches to choose one rather than both; in contrast if this is an attribute name, choosing both rather than one is a simple a union of two search results.

glguy:
Subclassing swcomponent won’t work well because the components aren’t changing just because someone did an extra analysis.

The fact that the control flow can be computed at the time of extraction is a bit of an accident here. While the same activity is generating both there’s no reason to assume that’s how things would always work. The tool ingesting software might well be different from the one doing control flow analysis, and there might be other analyses done using other tools requiring separate provenance tracking.

We could do extra subtypes instead of a flag on the subtype, but I would expect that knowing if something's conditional or not won't be a very important distinction in general. I'm happy to be wrong there, but making extra relations will make it harder to write queries to find thing that happen to have any control flow independent of conditions.

Other control flow subtypes might well contain details on what the conditions were.

glguy:
Also, the goal of making control flow it's own type with some subclassing ability was that control flow is a generic concept that's mentioned in DO-178C's objectives. We wanted to provide a generate way to capture the control flow while still allowing specialization like this case of conditional/unconditional or more detailed information about the conditions.

glguy:
Closed following the most recent DMDT discussion

PR #913, opened 2023-02-16
always generate posix paths in manifests

glguy:
I've adapted the manifest builder to always build paths with forward slash. My reading of the documentation is that Python on Windows will translate these automatically.

@kityansiu does this fix generating the turnstile example for you?

### DIFF
--- a/cli/rack/__init__.py
+++ b/cli/rack/__init__.py
@@ -22,7 +22,7 @@ from enum import Enum, unique
 from io import StringIO
 import logging
 from os import environ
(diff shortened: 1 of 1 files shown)


glguy:
Well shoot. I was hoping that by using the posix-specific path class it wouldn't auto detect and use windows path separators. Back to the drawing board <_<

glguy:
@weisenje after reading more documentation (and testing on a spare laptop) I think this commit is the fix

Issue #912, opened 2023-02-16
copy-to-default-graph : can we fix it before we really start using it


cuddihyge:
Copy-to-default-graph is also very Fuseki specific and always tied to optimization, which we are not doing during manifest loader.

Before there are any official manifests released that use this feature, could we change it to:

**copy-results-to-graph:   uri://DefaultGraph**    (  or "urn:x-arq:DefaultGraph" )
**perform-entity-resolution:  uri://DefaultGraph**

then the default graph could be in the footprint.

Callers could check the footprint for the default graph and fuseki and decide whether to optimize.

If/when we move to another triplestore that uses datasets differently, we could still copy all the results to a different graph for entity resolution, or other purposes.

- [x] change `copy-to-default-graph: true` to `copy-to-graph: uri://DefaultGraph`
- [x] change `perform-entity-resolution: true` to `perform-entity-resolution:  uri://DefaultGraph`
- [x] REMOVE `load(...boolean loadToDefaultGraph...)`  (in Python: `boolean default_graph`)
- [x] corresponding Python changes
(No released ingestion packages are using the above yet.)

cuddihyge:
@weisenje @glguy  comments?

There is a non-zero chance we'll actually need this functionality since Fuseki is getting slow and other solutions handle the default graph a little differently.

glguy:
I'm OK with this, but it does make ingestion packages triplestore-specific where before that tooling could be held responsible to know which default graph URI to use based on the runtime flag indicating which triplestore was being used.

cuddihyge:
**loadToDefaultGraph** could be changed to **loadtoGraph** and be a graph name instead of boolean.   Caller could specify uri://DefaultGraph ( or "urn:x-arq:DefaultGraph" ) if desired.

glguy:
If we're going to make it easy to bake fuseki conventions into the manifest file, then we should also add the triple-store-type field to the manifest file. Ingestion packages will become more specific to the triple store being loaded, so we'll need a way to capture that.

cuddihyge:
Do you mean **uri://DefaultGraph ( or "urn:x-arq:DefaultGraph" )** ?
Those are not fuseki conventions.  They are a w3c standard.   Fuseki interprets them differently.
I don't understand where you're going.

glguy:
You mentioned that we'd only want to copy to this particular graph on Fuseki and that other triple stores might want a different target.

PR #918, opened 2023-02-21
Copy-to-default-graph-generalization

glguy:
See #912

### DIFF
--- a/cli/rack/__init__.py
+++ b/cli/rack/__init__.py
@@ -528,9 +528,9 @@ def ingest_manifest_driver(
 
         if tmp_dir is not None:
             @with_status(f'Unpacking archive')
-            def go() -> None:
+            def unpack() -> None:
                 shutil.unpack_archive(manifest_path, tmp_dir)
-            go()
+            unpack()
 
             manifest_path = Path(tmp_dir) / 'manifest.yaml'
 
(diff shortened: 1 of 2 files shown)


cli/rack/manifest.py
```diff
@@ -13,7 +13,7 @@
         'name': {'type': 'string'},
         'description': {'type': 'string'},
 
-        'copy-to-default-graph':            {'type': 'boolean'},
+        'copy-to-graph':                    {'type': 'string'},
         'perform-entity-resolution':        {'type': 'boolean'},
```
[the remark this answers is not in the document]
glguy: I guess now's as good a time as any

cli/rack/manifest.py
```diff
@@ -13,7 +13,7 @@
...
         'perform-triplestore-optimization': {'type': 'boolean'},
```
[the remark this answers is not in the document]
glguy: sure

cli/rack/__init__.py
```diff
@@ -501,7 +500,7 @@ def build_manifest_driver(
     zipfile_path: Path
 ) -> None:
 
-    with tempfile.TemporaryDirectory() as outdir:
+    with TemporaryDirectory() as outdir:
```
Ptival, line 503: Is it the case that `builder.model_graphs` and `builder.data_graphs` depend on having created the archive?
At face value it looks like it's never explicitly used, but I assume there's a reason why it needs to be created, even if temporarily.

cli/rack/__init__.py
```diff
@@ -541,51 +548,46 @@ def ingest_manifest_driver(
...
-
-        if manifest.getPerformOptimization():
-            invoke_optimization(optimization_url)
+            if not manifest.getNodegroupsFootprint() == []:
+                delete_nodegroups_driver(manifest.getNodegroupsFootprint(), True, True, True, base_url)
+
+        for (step_type, step_data) in manifest.steps:
+            if StepType.DATA == step_type:
+                stepFile = base_path / step_data
+                ingest_data_driver(stepFile, base_url, None, None, triple_store, triple_store_type, False)
```
Ptival, line 557: `None, None` here mean that both the model graphs and the data graphs are the default graph?

Ptival (approved):
Looks fine. Left two questions to understand a bit of the context.

cli/rack/__init__.py
glguy: It means they aren't overridden from what's specified in their import.yaml files.

When you're using `rack data import` directly you can override those graph URLs with a CLI flag

cli/rack/__init__.py
glguy: I'm currently only collecting those graph names as debugging output. I am imagining that we could extend that in the future to report and error if the URLs were missing from the manifest's footprint:

Issue #951, opened 2023-03-28
Upgrade RACK's Fuseki to latest and RACK graph to TDB2


cuddihyge:

tuxji:
So, to confirm:

1. Bump Fuseki and Jena from 4.6.1 to 4.7.0 in the files below,

```rg
.github/workflows/assemble-files.yml
51:    - name: Download Fuseki
54:        curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-fuseki-4.6.1.tar.gz -o RACK/rack-box/files/fuseki.tar.gz
56:    - name: Download Apache Jena
59:        curl -LSfs https://archive.apache.org/dist/jena/binaries/apache-jena-4.6.1.tar.gz -o RACK/rack-box/files/jena.tar.gz
```

2. Switch from tdb to tdb2 in the files below,

```rg
cli/optimize.sh
18:# Currently using TDB1, this should become tdb2.tdbstats for TDB2
19:TDBSTATS="/opt/jena/bin/tdbstats"
26:if [[ ! -x "${TDBSTATS}" ]]; then
27:    echo "Aborting, ${TDBSTATS} is not executable"
31:echo "Running tdbstats"
34:"${TDBSTATS}" --desc "${RACK_FUSEKI_CONFIG}" > "${TMP_STATS}" && mv "${TMP_STATS}" "${STATS}"

rack-box/scripts/install.sh
155:curl -Ss -d 'dbName=RACK' -d 'dbType=tdb' 'http://localhost:3030/$/datasets'
```

Right?

PR #953, opened 2023-03-29
get the pid before stopping in optimize script

glguy:
Gets the PID before stopping fuseki so that it can tell if the stopping succeeded.

When I test this it successfully stops the service, however I'm getting an error from tdbstats that that doesn't seem related to my change.

```
Stopping Fuseki
Running tdbstats
org.apache.jena.sparql.ARQException: No such type: <http://jena.apache.org/2016/tdb#DatasetTDB2>
	at org.apache.jena.sparql.core.assembler.AssemblerUtils.build(AssemblerUtils.java:140)
	at org.apache.jena.sparql.core.assembler.AssemblerUtils.build(AssemblerUtils.java:132)
	at tdb2.cmdline.ModTDBDataset.createDataset(ModTDBDataset.java:81)
	at arq.cmdline.ModDataset.getDataset(ModDataset.java:35)
	at tdb2.cmdline.CmdTDB.getDataset(CmdTDB.java:77)
	at tdb2.cmdline.CmdTDB.getDatasetGraph(CmdTDB.java:69)
	at tdb2.tdbstats.exec(tdbstats.java:94)
	at org.apache.jena.cmd.CmdMain.mainMethod(CmdMain.java:87)
	at org.apache.jena.cmd.CmdMain.mainRun(CmdMain.java:56)
	at org.apache.jena.cmd.CmdMain.mainRun(CmdMain.java:43)
	at tdb2.tdbstats.main(tdbstats.java:43)
Now printing the contents of stats.opt:
cat: /etc/fuseki/databases/RACK/stats.opt: No such file or directory
```

### DIFF
--- a/cli/optimize.sh
+++ b/cli/optimize.sh
@@ -3,17 +3,20 @@
 
 set -eu
 
-echo "Stopping Fuseki"
-FUSEKI_PID=$(systemctl show --property MainPID fuseki)
-FUSEKI_PID=${FUSEKI_PID#"MainPID="}
-if [ -n "${FUSEKI_PID}" ]; then
+if systemctl is-active --quiet fuseki; then
+
+    echo "Stopping Fuseki"
+
+    # remember the PID in case stopping fuseki doesn't stop it    
+    FUSEKI_PID=$(systemctl show --property MainPID fuseki)
+    FUSEKI_PID=${FUSEKI_PID#"MainPID="}
+
     systemctl stop fuseki
-fi
-# systemctl stop doesn't always work, so kill -9 for now to be sure
-FUSEKI_PID=$(systemctl show --property MainPID fuseki)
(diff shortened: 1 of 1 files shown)


tuxji:
> This is most likely fallout from updating to TDB2 (#951) 

I agree.  Likely you tested with an old RACK still using tdb, but your cli/optimize.sh is already running tdb2.tdbstats.

glguy:
OK, that makes sense. I have to manually rebuild my rack docker image to get the aarch64 build, so I'm not always on the latest revision there. In that case I think this change is ready to go.

glguy:
We either need them in both places or neither places. It's possible that something has changed and fuseki restarts fine on its own now, but also it's easy for the restart script to report no error and not actually restart anything.

Issue #946, opened 2023-03-24
Add Turnstile zip as a GitHub release asset


weisenje:
Make a Turnstile ingestion package available as a GitHub release asset

- [x] create Turnstile ingestion package, add manually to V12 release
- [x] automate for future releases

kquick:
The portion that would generate the Turnstile Build and the associated data is here: https://github.com/ge-high-assurance/RACK/blob/master/Turnstile-Example/Turnstile-IngestionPackage/Load-TurnstileData.sh#L100-L109

The only change needed would be to add `-o {OWLFILE}` to the `ingest_data` command at the end.

I don't think we want to change the `Load-TurnstileData.sh` file itself, but I'm not sure where this work is being done, so I can identify this as what is needed for wherever we are scripting this artifact-generation process.

tuxji:
> I don't think we want to change the Load-TurnstileData.sh file itself, but I'm not sure where this work is being done, so I can identify this as what is needed for wherever we are scripting this artifact-generation process.

We script the artifact-generation process in this [.github/workflows/release.yml](https://github.com/ge-high-assurance/RACK/blob/master/.github/workflows/release.yml) file.  I don't suppose we want the Turnstile model, nodegroups, and data to be loaded into the RACK box, so we would NOT modify the `.github/workflows/assemble-files.yml`, `.github/workflows/build-docker-image.yml`, `.github/workflows/build-virtual-machine.yml`, or `rack-box/scripts/install.sh` files.  Instead, I suppose we want to append a new job to the bottom of `.github/workflows/release.yml` which will start up the newly built rack-box docker image, run `cli/setup-turnstile.sh` to populate the rack-box with the Turnstile model and nodegroups, run `Turnstile-Example/Turnstile-IngestionPackage/Load-TurnstileData.sh` to populate the rack-box with the Turnstile data, run whatever additional step is needed to generate the ingestion package, kill the rack-box container since we no longer need it, and upload the ingestion package as a GitHub release artifact.  Does that sound correct?

tuxji:
OK, let's rename the `Load-TurnstileData.sh` script to a better name like `Assemble-TurnstileData.sh` (pick your choice of assemble, generate, etc.).  Then the new CI job only needs to clone the RACK repository, call that script to create `turnstile-verXYZ.zip`, and attach the ingestion package to the release page.  Right?

kquick:
I think that's generally right.  Here's more details as I see them, but I'd want others to confirm this:

[...]

PR #960, opened 2023-04-04
Automate construction of complete Turnstile ingestion package

glguy:
See #946

Remaining work: incorporate into CI


glguy:
@tuxji I think what's left is to invoke this script from the CI process and save the zip file output somewhere.

kquick (approved):

tuxji:
Looks good.  I'm traveling on Wednesday, so I won't be able to make the CI changes until Thursday.

tuxji:
@glguy @kquick Please review the [commit](https://github.com/ge-high-assurance/RACK/pull/960/commits/197cb72f42727022da5cdf81c50b266f3c9a898f) I just added to this pull request (I also rebased it on the master branch to bring it up to date).  I've modified the release workflow to build the Turnstile ingestion package and attach it to the release. I've also taken advantage of the fact that we're now using Ubuntu 22.04 which already has swi-prolog version 8.4.2 to simplify all the places which had to add swi-prolog's ppa repository before.  I figure we can remove these extra steps now as long as ASSIST doesn't need any features that are only in SWI-Prolog version 9.0.4 (the ppa version).

tuxji:
P.S. I ran the workflow's new commands manually in my checkout so I think they should work on the github runner.  However, you should review these [commands](https://github.com/ge-high-assurance/RACK/blob/issue946/.github/workflows/assemble-turnstile-data.yml) as well.  I have to wait until we merge this pull request into the master branch before the Actions tab will let me call the assemble-turnstile-data.yml workflow from the Actions tab.  If it works when I call it from the Actions tab for testing, it should also work when CI calls it for the next release.

.github/workflows/assemble-turnstile-data.yml
```diff
@@ -0,0 +1,61 @@
+# Assembles Turnstile ingestion package (reusable workflow)
+
+on:
+  workflow_call:
```
kquick, line 4: I don't think `workflow_call` should have a trailing colon (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call)

kquick (approved):
Looks good, and it's nice not to have to do extra work to get SWI Prolog available.

I noticed that this new action isn't run on the branch... I suppose we have  to wait until merge time to see if it works OK?

kquick:
> I noticed that this new action isn't run on the branch... I suppose we have to wait until merge time to see if it works OK?

Ah, I re-read your P.S. comment after making this remark... just ignore it.  :-)

.github/workflows/assemble-turnstile-data.yml
tuxji: Yes, you're correct.  I'd copied this file from build-virtual-machine.yml where the `workflow_call` had been followed by an input, and I'd deleted the input but not the trailing colon.  I'll go ahead, merge the pull request, and if CI fails immediately, I'll fix it, otherwise I'll wait for CI to find any other problems that I can fix at the same time too.

PR #962, opened 2023-04-06
Organize all the overlays together

glguy:
This changes all overlays to be top-level packages that can be included by referencing a single, overlay-top-level manifest.

A new ARCOS overlay is now defined that pulls in all the overlays needed for the demonstration application.

We'll want to set up Eclipse .projects for all of these, I think, but this change doesn't do that yet. Some of the .projects already exist.

The primary driver for this change is to support the much more modular way we're now building ingestion packages using RITE.


Ptival (commented):
On top of my other remark, just wondering whether it is intended that the `overlays/MITRE-CWE` directory has a copy of some `gen-capec.py` as well as the seemingly-generated `CAPEC.csv` and `CAPEC_CWE.csv`, while in the `overlays/CAPEC` directory, there is another copy of the script, but no generated files checked in.

Otherwise everything looks reasonable.

PR #985, opened 2023-07-26
RACK-928: Slow Startup RiB Container on Apple M1

chrisage:
RACK-928: Slow Startup RiB Container on Apple M1

Enhancing github workflow to include AARCH64 (Apple Silicon) platform docker image. The linux/amd64 os/arch images were not performant on M1 machines.

Solves: RACK-928


chrisage:
Requires M1 Runner (or qemu emulation through buildx)

Issue #986, opened 2023-08-03
Make SemTK cardinality checks accessible by RACK tools


weisenje:
**Background**:
Users can access cardinality restriction violations (instances with too few/many of a given property) in the SemTK SPARQLgraph UI.  We now want to make this violation information accessible to other tools in the RACK ecosystem.

**Acceptance Criteria**:
- [x] add SemTK Java client call for future use by RITE.   **_DONE: OntologyInfoClient.execGetCardinalityViolations()_**
- [x] add semtk-python3 call for use by RACK CLI.  **_DONE: semtk3.get_cardinality_violations()_**
- [x] add RACK CLI call.  **_DONE: (https://github.com/ge-high-assurance/RACK/pull/988)_**
- [x] add RACK UI button to directly open SPARQLgraph Explore restriction mode.  _**DONE: (https://github.com/ge-high-assurance/RACK/pull/990)**_
- [x] document on [semtk-python3 wiki](https://semtk-python3.readthedocs.io/en/latest/api/semtk3.html)
- [x] document on [RACK CLI wiki](https://github.com/ge-high-assurance/RACK/wiki/RACK-CLI) - @glguy

glguy:
I'll jump on it

PR #988, opened 2023-08-04
Expose cardinality checker via the cli

glguy:
See #986

### DIFF
--- a/cli/rack/__init__.py
+++ b/cli/rack/__init__.py
@@ -565,7 +565,9 @@ def cardinality_driver(
         triple_store: Optional[Url],
         triple_store_type: Optional[str],
         headers: bool,
-        export_format: ExportFormat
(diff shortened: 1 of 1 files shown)


saratge:
I am getting the error shown below after ingesting using this ingestion package: [gsn_demo.zip](https://github.com/ge-high-assurance/RACK/files/12264983/gsn_demo.zip)

image

The error appears to arise from python semtk client

cuddihyge:
The cardinality bug has to do with @saratge 's use of unions inside SADL statements, which are correct, but no one had used before with SemTK.
I can fix it on the master branch.
This bug has nothing to do with @glguy 's work on this branch.
I'm fine merging this branch.
The bug will persist until I make another semtk release and add it to RACK.

cuddihyge:
I opened #989 to address the bug.

PR #993, opened 2023-08-21
Modernize the RACK cli packaging

glguy:


glguy:
@kityansiu this should address the messages you were getting about using a deprecated install process. The process is now very modern!

glguy:
@tuxji would you have time to check that I didn't do something silly to the Github CI process?

.github/workflows/continuous.yml
```diff
@@ -29,19 +29,21 @@ jobs:
...
+        key: ${{ runner.os }}-pip-${{ hashFiles('cli/pyproject.toml') }}
         restore-keys: |
           ${{ runner.os }}-pip-
           ${{ runner.os }}-
 
     - name: Install RACK CLI development dependencies
-      run: pip3 install -q -r cli/dev/requirements.txt
+      run: |
+        cd cli
+        pip3 install ".[dev]"
```
tuxji, line 40: I'm not familiar with this ".[dev]".  What does it mean?

.github/workflows/continuous.yml
```diff
@@ -108,9 +110,9 @@ jobs:
 
     - name: Run rack-box tests
       run: |
-        cd RACK
-        pip3 install -r cli/requirements.txt
-        pip3 install cli/.
+        cd RACK/cli
+        pip3 install .
+        cd ..
```
tuxji, line 115: Too bad that you need two cd commands.  What makes `pip3 install ./cli` no longer work?

cli/README.md
```diff
@@ -46,17 +46,15 @@ the RACK-Box README.
 ```shell
 python3 -m venv venv
 source venv/bin/activate
-pip install --force -r requirements.txt
-python3 setup.py install
+pip install --force-reinstall .
```
tuxji, line 49: Don't you need `pip3`?

cli/README.md
```diff
@@ -46,17 +46,15 @@ the RACK-Box README.
...
 ```
 
 If running on Windows, [GitBash](https://gitforwindows.org/) can be used with the following commands.
 
 ```shell
 python -m venv venv
 source venv/Scripts/activate
-pip install --force -r requirements.txt
-python setup.py install
+pip install --force-reinstall .
```
tuxji, line 57: Shouldn't `pip` be `pip3`?

cli/pyproject.toml
```diff
@@ -0,0 +1,48 @@
...
+optional-dependencies.dev = [
+    "mypy==1.0.0",
+    "pylint==2.16.2",
+    "pytest-docker==1.0.1",
+    "pytest==7.2.1",
+    "typed-ast==1.5.4",
+    "types-PyYAML==6.0.12.6",
+    "types-requests==2.28.11.13",
+    "types-tabulate==0.9.0.0",
+]
```
tuxji, line 42: OK, I assume that `pip3 install .` and `pip3 install ".[dev]"` install these `dependencies` and `optional-dependencies.dev` respectively.

rack-box/scripts/install.sh
```diff
@@ -183,5 +183,5 @@ done
 # Setup the RACK dataset using the RACK CLI
 
 cd "/home/${USER}/RACK/cli/"
-python3 -m pip install ./wheels/*.whl
+python3 -m pip install --no-dependencies ./wheels/*.whl
```
tuxji, line 186: Why not `pip3 install`?

tuxji (approved):
+1

Looks good!  Just have a few questions to better understand what's going on.

.github/workflows/continuous.yml
glguy: This is (kind of concerning) pip syntax that means to install the package in the current directory `.` using the optional dependency set `dev`. Another option was for me to write `rack[dev]`

cli/README.md
glguy: The environments I've been working in *only* have pip version 3, and pip points to that. I'll change the documentation to be explicit about being version 3 for platforms where version 2 still exists

rack-box/scripts/install.sh
glguy: only because I wasn't thinking about changing this line of code at all until it failed due to dependency checks. I was singularly focused on `--no-dependencies` when I visited this line. I'll make the change.

In case you wondered and didn't consult the failed build logs, the `--no-dependencies` became necessary because pip was able to tell that the version of `python3-semtk` that we'd installed was not the GitHub version, as specified in the `pyproject.toml` but it was the pre-compiled wheel file which self-descripted as version `0.1`. This flag disables all the checks. I believe that this is OK because all the checks were done at the point at which we generated the wheels.

cli/pyproject.toml
glguy: You're quite right!

.github/workflows/continuous.yml
glguy: First I tried what you wrote, and that didn't work. I'm new to using `pyproject.toml` and when I found that the code committed above working, I went with it.

Your question prompted me to *try harder* and I learned that a trailing `/` indicates to pip that what I want is for it to install a directory, rather than a package name. I've updated the script to use a single command and saved the extra `cd` for another day!

PR #998, opened 2023-08-31
disable dependency checking in assemble-turnstile-data

glguy:
This disables dependency checking when installing the wheel files. the dependency checking will have already been done at this point. Git versions of things seem to break normal wheel installation.


glguy:
@tuxji I think this is the first. I'll watch CI

tuxji (approved):
+1

Thanks!  CI runs the assemble-turnstile-data workflow only when we build a release, but we'll test it tomorrow.

PR #1002, opened 2023-09-05
migration: v13 changes

Ptival:
I have created some helpers to deal with some very repetitive changes.

Unfortunately, right now this does not support disjunctive types (e.g. {NS1:Class1 or NS2:Class2}), so I wrote slightly incorrect descriptions in the couple occurrences of those.

It'd be a bit more work to add proper support.


glguy (approved):
The only obvious thing I see is that we don't have a way to explain the constraints around `IgnoredFailureOOB` but other than that I think this is great and we should get it out and onto the wiki

